PR #54, opened 2026-09-03
Migrate watchdog: supervision loop + systemd/SCM service install

### LINKED ISSUE #9: Migrate watchdog
**Component:** `watchdog/` · **Source:** migrate from `old/watchdog`

Process supervision: restart agent on crash/kill, binary integrity check, tamper event reporting. Linux (systemd interaction) first; Windows SCM service arrives with M3.

## Done when
- [ ] Kill-the-agent scenario: watchdog restarts it and records the event
- [ ] Runs as the systemd unit packaging will ship

florianamette:
Closes #9

The last M1 item. Ported from `old/watchdog` with the agent's new name and flags (`agent run --alerts`), the systemd unit renamed to `synthaea-agent.service`, and the Windows SCM machinery (service self-dispatch with the 1063 fallback, `sc failure` triple-restart policy) preserved verbatim for the M3 return.

Worth reviewing: one latent bug fixed — the supervised child's log redirect was a hardcoded `C:\Windows\Temp` path in the *shared* loop, which on Linux created a literal file named `C:\Windows\Temp\...` in the working directory. Now per-OS (`/var/tmp/synthaea-agent.log` on unix).

## Test plan

- [x] 3 unit tests (path resolution, UNC stripping); clippy `-D warnings`, fmt, check-deps clean
- [ ] Live kill→restart validation on ubuntu2404 — running now, results will be posted here


florianamette:
## Live VM validation: PASS

ubuntu2404, supervision chain `watchdog run` → `agent run`:

```
[watchdog] starting the agent...
[watchdog] agent exited (signal: 9 (SIGKILL)). Restarting in 5s...
[watchdog] starting the agent...
PASS: agent respawned (pid 1417 -> 1480, starts=2)
```

SIGKILL on the agent was detected and the agent respawned after the 5s delay, exactly per design. (A first automated run reported FAIL — that was the test harness's own `pkill -f` pattern matching the ssh shell's command line, not a watchdog defect; the by-pid rerun above is conclusive.)

Issue #53, opened 2026-09-03
eBPF read_ppid returns garbage on kernels other than the bindings' source kernel (no CO-RE)


florianamette:
**Component:** `crates/sensors/linux/ebpf` · found during the #52 walking-skeleton VM validation

On ubuntu2404 (6.8.0-63-generic, aarch64), self-spawn alerts report `ppid=4294901760` (0xFFFF0000). The aarch64 `vmlinux_aarch64.rs` bindings were generated from a 6.17 Oracle kernel; `read_ppid` walks `task_struct.real_parent->tgid` by frozen offset, so any kernel with a different `task_struct` layout yields garbage. The probes' own docs call this out as the no-CO-RE limitation; the lab matrix caught it on its second-ever run.

Impact: every ppid-dependent rule (web-server lineage, PARENT-SUSPECT, LOLBIN, SELF-SPAWN parent exclusions) is unreliable off the binding kernel. Beacon and stateless rules are unaffected.

Options, in rough preference order:
1. **Adopt CO-RE** (`aya_ebpf::bpf_core_read!` / relocations) for `task_struct` reads — the kernels in MATRIX.md all ship BTF, which is the prerequisite; this removes the per-kernel binding freeze entirely.
2. Generate bindings per matrix kernel and select at load time (heavy, scales badly).
3. Read ppid in userspace via /proc as a fallback when the kernel value is implausible (racy; the sensor doc explains why exec-time capture matters).

## Done when
- [ ] ppid is correct across the Linux rows of MATRIX.md (asserted by a scenario or conformance check that spawns a known parent/child pair)

PR #155, opened 2026-09-09
sensor-linux: read argv from /proc, drop the last frozen offset (#152)

### LINKED ISSUE #152: eBPF probe: portable argv/cmdline (last frozen-offset read)
Follow-up from #149 (Closes #53, #111).

After #149, the Linux eBPF `sched_process_exec` handler has **one remaining
frozen-offset read**: `read_argv()` walks `task_struct -> mm_struct ->
{arg_start, arg_end}` via the `vmlinux*` bindings (and the `i686_offsets`
module on 32-bit). Correct only on the kernel the bindings were generated from;
a short or empty `cmdline` elsewhere — never a crash, and `image_path` / `ppid`
/ `parent_comm` are all portable now, so this is the last piece.

## Options

1. **Tracepoint fields** — `sched_process_exec` does not carry argv. The
   `sys_enter_execve` tracepoint has `argv` (a `char **`), but it fires
   *before* the exec commits (failed execs included) and reading the vector
   needs a bounded per-arg `bpf_probe_read_user_str` loop — the same verifier
   budget the #149 work just fought. Would also re-add the tracepoint #149
   dropped.
2. **`bpf_probe_read_kernel` of `mm->arg_start/arg_end` via CO-RE** — aya still
   has no CO-RE (aya-rs/aya#349); blocked on the same thing as the original
   #53 ppid read (which #149 sidestepped with a tracepoint-fed map).
3. **Per-CPU scratch + `bpf_probe_read_user`** of the `arg_start..arg_end`
   blob, offsets read via CO-RE once aya#349 lands.
4. **Userspace `/proc/<pid>/cmdline`** on the exec event — racy for
   short-lived processes (the reason #149 kept the image path in the probe),
   but `cmdline` is display/analysis-only, never an identity input, so the
   race is more tolerable here.

## Done when

`cmdline` / `argv` are correct across the Linux rows of `lab/MATRIX.md`
(asserted by a scenario or conformance check), with no `task_struct` /
`mm_struct` frozen-offset read left in `crates/sensors/linux/ebpf`.

old-dov:
Closes #152. Follow-up to #149/#111/#53 — the eBPF probe now reads **no** `task_struct`/`mm_struct` frozen offset at all.

## The last offset
`read_argv` walked `task_struct → mm_struct → {arg_start, arg_end}` via the `vmlinux*` bindings (and a raw `i686_offsets` table on 32-bit) — correct on the kernel the bindings came from, a short/empty `cmdline` anywhere else. It was the one thing keeping the per-kernel binding files alive.

## Move to userspace (issue option 4)
The loader reads `/proc/<pid>/cmdline` when it drains the exec event. That blob *is* the NUL-separated argv, so parsing is trivial and there's no product-side length cap any more.

- **ebpf**: delete `read_argv` (×3 arch), `i686_offsets`, and the `mod vmlinux` / `mod vmlinux_aarch64` declarations — which lets the two generated binding files go too (**−120k lines**). The probe now reads only stable tracepoint fields and syscall arguments.
- **wire**: `ExecEvent` drops `cmdline`/`cmdline_len`; `MAX_CMDLINE_LEN` gone; `WIRE_VERSION` 2 → 3.
- **userspace**: `read_proc_cmdline(pid)` + a pure, tested `parse_proc_cmdline`; `normalize::exec` takes `argv` from the caller, keeps `cmdline` as the space-joined render for Sigma. `drain!` now takes `Fn(&wire) -> Event` so only the exec path pays the `/proc` read.
- **lab**: `scenarios/argv.sh` asserts capture per kernel (the T1059.004 base64 rule fires iff `cmdline` was populated); `MATRIX.md` updated — no frozen offset, no offset table to regenerate.

## Tradeoff
A process that exits within the ~ms drain latency leaves no `/proc` entry → empty argv; a reused pid in that window would mis-attribute argv. Accepted: `argv`/`cmdline` are analysis inputs, never identity — `image_path` is still read authoritatively in the probe. Documented in `read_proc_cmdline` and `argv.sh`.

## Verification
`cargo build -p sensor-linux` (ebpf compiles for `bpfel-unknown-none`, bpf-linker present) · sensor tests (16, incl. 5 new `parse_proc_cmdline`) · `cargo clippy --workspace --exclude sensor-linux-ebpf --all-targets -- -D warnings` · `tools/check-deps.py` (38 crates). CI billing is off — run locally.

Needs real-hardware validation on the `MATRIX.md` kernel rows (`argv.sh`) — 5.15 / 6.1 / 6.8 / 6.12 — before trusting outside the lab.


old-dov:
Synced with `main` (`98bfe72` merges #153/#154 + the Cargo.toml conflict fix `fa30620`). Re-verified locally against current `main`:

- `cargo build -p sensor-linux` — eBPF cross-compile OK (bpf-linker)
- `cargo test -p sensor-linux` — 16 pass
- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --exclude sensor-linux-ebpf --all-targets -- -D warnings` — clean
- `cargo test -p ml -p schema` — all pass
- `tools/check-deps.py` — 38 crates OK

No conflicts; `sensor-linux-wire` (WIRE_VERSION 2→3) is isolated — nothing outside `crates/sensors/linux/` references wire types.

Still pending: real-hardware validation on the `lab/MATRIX.md` kernel rows (5.15 / 6.1 / 6.8 / 6.12) via `lab/scenarios/argv.sh`.

old-dov:
## Hardware validation — kernels 5.15 and 6.1

Ran on Hyper-V VMs (`lab/vagrant-hyperv/`), branch built from source in-VM (`cargo build --release -p agent -p sensor-linux`).

| | 5.15.0-91-generic (Ubuntu 22.04) | 6.1.0-17-amd64 (Debian 12) |
|---|---|---|
| `agent status` — verifier | 5/5 programs accepted | 5/5 programs accepted |
| `lab/scenarios/lineage.sh` (#53/#111) | 3/3 `T1059 … ppid=<nginx> comm=nginx` | 3/3 idem |
| `lab/scenarios/argv.sh` (#152) | 3/3 `T1059.004 … base64 decode` | 3/3 idem |

**Lineage** — every `sh` exec event carries the correct `ppid` (the `nginx` pid) and `parent_comm="nginx"`. No `ppid=0`, no `4294901760`. `image_path` is the real path (`/bin/sh`, `/tmp/nginx`), not `argv[0]`.

**argv** — every `sh -c` exec event carries the full vector from `/proc/<pid>/cmdline`:
`argv=["sh","-c","echo ZWNobyBoZWxsbwo= | base64 -d >/dev/null"]` — three tokens, interior spaces preserved, never empty. `cmdline` is the space-joined form for Sigma.

5.15 is the strictest verifier of the four matrix rows, so these two cover most kernel-portability risk. **6.8 / 6.12 still pending** — no ready Hyper-V box (`lab/vagrant-hyperv/README.md` § Missing boxes); needs a built box or a run on native Debian 13 / Ubuntu 24.04.

### Unrelated finding (not a regression, not from this PR)

Both scenarios also fire `T1059 — comm=<x> spawned 3x in 30s by ppid=<parent> — suspected self-spawn` on the loop bodies (`nginx`/`sh`/`sleep`). `ppid` is correct here (the scenario's bash loop), so this is the self-spawn heuristic being over-eager on a short shell loop — the same false positive class reported in the Gafgyt analysis. Worth a separate calibration issue; flagging here since these scenarios are a clean repro.

Sollykhan:
Findings on WSL2 6.6 : read_proc_cmdline returns ENOENT for every exec event, including long-lived processes like sleep 120. Root cause: the tokio select! loop drains the ring buffer in batches with enough latency that even multi-second processes have exited by the time /proc/<pid>/cmdline is read.

The race is inherent to the design. Two mitigation paths:

Dedicated drain thread — move the ring buffer poll off the tokio runtime into a bare std::thread that spins on rb.next() and reads /proc immediately after each event. Reduces the race window from ~tens of ms to ~microseconds.
Accept the limitation — the code comment already documents this. For the EDR use case (malware processes live longer), it may be acceptable with a note in the PR.

The /proc/1/root/proc fallback I proposed earlier doesn't help here — it's not a namespace issue (NSpid = single value = root PID namespace confirmed). It's purely timing.

old-dov:
## Hardware validation — Debian-family x86_64 rows

Ran on the Hyper-V harness (`lab/vagrant-hyperv/`), `lab/validate-155.sh` per kernel:

| Row | Kernel | verifier | argv (#152) | lineage (#53) | unit tests |
| --- | --- | --- | --- | --- | --- |
| `ubuntu2204` | 5.15.0-91 | 5/5 accepted | ≥3 `T1059.004`, `sh -c` full argv | ≥3 `T1059`, no `ppid=0` | 16 pass |
| `debian12` | 6.1.0-17 | 5/5 accepted | ≥3 `T1059.004`, `sh -c` full argv | ≥3 `T1059`, no `ppid=0` | 16 pass |

Both `[PASS]`. The eBPF object loads and every program is verifier-accepted on 5.15 (the strictest of the four), argv is captured from `/proc/<pid>/cmdline` (the `T1059.004` cmdline-substring rule fires, which it only can if argv was populated), and the #53 parent-lineage alert still fires with no `ppid=0` regression.

The accepted userspace-read race is visible and harmless: 0–2 exec events per run with `argv:[]`, always a short-lived process that exited before the drain (e.g. the scenario's own `$(uname -r)`), never the cooperative `sh -c`.

**6.8 / 6.12 still pending** — no usable Hyper-V box (`generic/*` gap; `boxen/ubuntu-24.04` was tried and rejected — SMB `/vagrant` prompt + root disk too small for the toolchain). A home-built box or the arm64 `../vagrant` harness are the remaining paths.

### Also on the branch now
- `lab/vagrant-hyperv/` harness brought up to working order: `helpers.ps1` (dot-sourced shortcuts, base64-safe remote commands), Dynamic Memory, `.gitattributes` (LF on `*.sh`), `provisioning/linux-toolchain.sh` hardening (retry the bpf-linker fetch, pre-install the pinned toolchain, bust a stale probe-less build).
- `lab/validate-155.sh` + `lab/vagrant-hyperv/RUNBOOK-155.md` + `lab/lint.sh` + `lab/diag.sh`.
- Doc polish on `read_proc_cmdline` (the argv-spoofing gap: `/proc` is read at drain time, not execve) and a distinguished errno.

### Heads-up, not from this PR
`cargo fmt --all --check` is dirty on the branch — stable flags `engine.rs` / `policy` wrapping, `cargo +nightly fmt` wants import-grouping across ~25 files workspace-wide (`rustfmt.toml` sets nightly-only options the stable CI step can't enforce). Pre-existing, from the #153/#154 merge — wants its own `chore: nightly fmt` pass.

old-dov:
@Sollykhan d'accord sur ta conclusion. WSL2 6.6 : `sleep 60` avec parent confirmé → `ENOENT` quand même, donc pas la race de drain — c'est bien la couche interop WSL entre le contexte eBPF et la vue `/proc` de l'agent. WSL2 n'est pas une row de `lab/MATRIX.md`.

Gravé en note sur `read_proc_cmdline` dans `8ebca40` pour que ce ne soit pas re-creusé.

La branche a aussi pris le commentaire de validation hardware (5.15 + 6.1 PASS) et la remise en état du harness `lab/vagrant-hyperv/` (helpers, Dynamic Memory, `.gitattributes`, durcissement provisioning) + `validate-155.sh` / RUNBOOK / lint.sh / diag.sh.

Prêt à merger de mon côté. 6.8 / 6.12 restent pending (pas de box Hyper-V utilisable — `boxen/ubuntu-24.04` testée et rejetée), à faire hors #155.

old-dov:
## Self-review pass + re-validation (5.15 / 6.1)

Two small commits on the branch, both re-run on the Hyper-V harness (`lab/validate-155.sh` per kernel).

### `6537aa5` — sensor-linux: address self-review nits
- `read_proc_cmdline`: the exit-race arm only matched `ENOENT` (`ErrorKind::NotFound`); the `ESRCH` the comment claims is silenced maps to `ErrorKind::Uncategorized` and fell through to the `debug!` arm. Extracted `is_proc_exit_race()` (checks the raw errno too), with a regression test `proc_exit_race_is_silent_for_enoent_and_esrch_only`.
- Drain loop: softened the "no I/O wait" claim on the per-exec `/proc` read — it can still stall on `mmap_lock` contention or a page fault in the target's mm.
- `lab/scenarios/argv.sh`: marked executable (100644 → 100755) like every sibling scenario.

### `df8ca5c` — lab: hold the argv.sh cooperative sh past the drain latency
`sh -c "echo … | base64 -d >/dev/null"` is gone in ~1-2 ms. On 6.1 under load (a `debian-sa1` cron firing mid-run) the first iteration's exec event consistently reached the drain *after* the process had exited → empty `/proc/<pid>/cmdline` → `validate-155.sh`'s `sh -c full argv >= 3` failed on the **accepted userspace-read race**, not a regression. Prefixing the cooperative invocation with `sleep 0.2` holds the shell open well past the ~ms drain; the race keeps its own informational line. `lab/` only, no product code.

### Re-validation

| Row | Kernel | verifier | `cargo test -p sensor-linux` | argv (#152) | lineage (#53) | `validate-155.sh` |
| --- | --- | --- | --- | --- | --- | --- |
| `ubuntu2204` | 5.15.0-91 | 5/5 accepted | 17/17 | 3/3 `sh -c` full argv, `T1059.004` ×6 | ×5 `T1059`, 0 `ppid=0` | `[PASS]` |
| `debian12` | 6.1.0-17 | 5/5 accepted | 17/17 | 3/3 `sh -c` full argv, `T1059.004` ×6 | ×5 `T1059`, 0 `ppid=0` | `[PASS]` |

`cargo clippy -p sensor-linux --all-targets -- -D warnings` clean on 5.15 (same toolchain/code on 6.1). The accepted race is still visible and harmless — 0–1 exec events per run with `argv:[]`, always a short-lived pipeline child, never the `sleep`-held `sh -c`.

6.8 / 6.12 remain pending, out of this PR.

### Filed
- #159 — the over-eager SELF-SPAWN (T1059) alert on the scenario shell loops that `validate-155.sh` calls out as "known non-blocking noise". Pre-existing (Windows-calibrated rule, no Linux calibration), tracked separately.

Sollykhan:
## Hardware validation

Ran `lab/validate-155.sh` and `lab/scenarios/argv.sh` on:

- **Hyper-V, kernel 5.15** — ✅ argv populated end-to-end, T1059.004 fires on the base64 canary
- **Hyper-V, kernel 6.1** — ✅ same result

Both `MATRIX.md` rows for these kernels can be marked green. Haven't run 6.8 / 6.12 yet.

## WSL2 edge case (out of scope, worth noting)

Followed up on the "argv sometimes empty on WSL2" hunch — ran `sleep 60 &` and dumped `/proc/<pid>/cmdline` while the process was still alive (well past any drain-latency window):

Sollykhan (approved):
Zero-frozen-offset probe is a real architectural cleanup, not just a fix — the userspace /proc/cmdline read plus wire v3 tradeoff (argv/cmdline as analysis input, image_path as identity) reads clean. −120k lines of vmlinux bindings gone is a bonus. Hardware validation for 5.15 / 6.1 posted above; 6.8 / 6.12 still to run. Needs rebase + cargo +nightly fmt after #160 lands.

old-dov:
Rebased on `main` after #160 (nightly rustfmt) landed — no drift, `.gitattributes` add/add conflict from #160 vs the argv.sh LF fix merged cleanly (both additive).

- `cargo +nightly fmt --all --check` — clean on `crates/sensors/linux/`, no changes needed (the workspace-wide fmt gap in `agent/src/health.rs` + `crates/transport/*` from #157/#158 is unrelated to this PR, flagged separately)
- `cargo test -p sensor-linux` — 17/17
- `cargo clippy -p sensor-linux -p ml -p schema --all-targets -- -D warnings` — clean

Still pending per the review above: hardware validation on 6.8/6.12.

Sollykhan:
Nice rebase and self-review. WSL2 note baked into read_proc_cmdline is the right call — future maintainer won't chase it again. Approve still stands, merging.

PR #156, opened 2026-09-10
fix(store): two-phase drain/ack and active-segment cap enforcement

### LINKED ISSUE #108: EventSpool: two-phase drain/ack and active-segment cap enforcement
Review findings on the store migration PR:

1. **At-least-once gap**: drain deletes a segment before the consumer acknowledges the upload — a crash between delete and successful send loses the batch. Needs a two-phase protocol: hand out the segment, delete only on explicit ack; re-drain on restart.
2. **Active-segment cap bypass**: the byte cap sheds only sealed segments — the currently-open segment can grow past the cap unchecked. Enforce the cap including the active segment (seal early and shed).

Both in crates/store/src/spool.rs.

SherlockOmss:
## Summary

Fixes #108

- **Two-phase drain/ack protocol**: `drain_oldest()` no longer deletes segments immediately. Instead, it renames them to `.inflight` and requires an explicit `ack()` call after successful upload. On crash, `open()` recovers `.inflight` files for re-delivery (at-least-once semantics).

- **Active-segment cap enforcement**: `enforce_cap()` now handles the case where only the active segment exists and exceeds the byte cap. It seals the segment (rotates to a new one) before shedding, ensuring the cap is always enforced.

## Changes

| File | Description |
|------|-------------|
| `crates/store/src/spool.rs` | +168/-21 lines |

### New API
- `ack() -> io::Result<bool>`: Confirms successful upload, deletes `.inflight` file

### Modified behavior
- `drain_oldest()`: Renames to `.inflight` instead of deleting; idempotent if called multiple times without ack
- `open()`: Recovers `.inflight` files from previous crash
- `enforce_cap()`: Seals active segment if it's the only one and exceeds cap

## Test plan

- [x] `cargo check -p store`
- [x] `cargo clippy -p store -- -D warnings`
- [x] `cargo fmt -p store`
- [x] `cargo check --workspace --exclude sensor-linux-ebpf`
- [x] `tools/check-deps.py`

### New tests added
- `two_phase_drain_redelivers_on_crash`: Verifies segment recovery after crash
- `idempotent_drain_without_ack`: Verifies repeated drain returns same data
- `active_segment_cap_enforcement`: Verifies cap is enforced on active segment

🤖 Generated with [Claude Code](https://claude.ai/code)


old-dov:
Review pass (level `high`). No non-test callers of `EventSpool` yet, so this is an API-design + logic review. The two-phase rename-to-`.inflight` + `ack()` shape is the right idea, but a few gaps let the un-acked data the protocol exists to protect get dropped or stuck.

## Blocking

**1. A recovered `.inflight` segment is no longer cap-protected — `src/spool.rs:236`**
Drain segment 0, crash before ack. `open()` renames `spool-0.inflight` → `spool-0.jsonl`. A burst of `push()` before the uploader's next `drain_oldest` runs `enforce_cap()`, which sees `total > max_bytes` and — since segment 0 is now an ordinary oldest `.jsonl` with `in_flight == None` — `fs::remove_file`s it. The recovered, never-uploaded records are lost. Recovery silently fails under exactly the backpressure that makes the spool matter.

**2. No forward-progress guarantee — `src/spool.rs:132`**
`drain_oldest` hands out segment N; the server rejects that batch non-retryably (oversized / malformed / persistent 4xx), so `ack()` is never called. Every subsequent `drain_oldest` re-delivers segment N (`in_flight` stays `Some(N)`). Newer events pile into later segments and are shed by `enforce_cap`. The agent stops delivering all telemetry indefinitely, with no escape hatch — no max-retry, no poison-segment skip. The old delete-on-drain always made forward progress; this needs an explicit "give up on segment N after k attempts" path.

**3. Sealing + shedding the sole active segment discards the freshest events wholesale — `src/spool.rs:223`**
If `max_bytes` is smaller than a typical segment (misconfig, or large events), each `push` that tips `total` over the cap seals the active segment (`head_seq += 1`) and then deletes the whole file in the same `enforce_cap` loop, including the just-written records. The spool oscillates and retains almost nothing. Rotating on a byte threshold in `push()` — so no single segment can exceed the cap — would avoid special-casing the active segment here.

## Non-blocking

**4. `enforce_cap()` and `stats()` ignore the in-flight segment's bytes — `src/spool.rs:208`**
After `drain_oldest`, segment 0 is `spool-0.inflight` and invisible to `segment_seqs()`. While the caller retries a slow upload, `push()` keeps writing new `.jsonl` segments; `enforce_cap` only sums `.jsonl`, so real disk usage can reach `max_bytes` + one full segment. `stats().bytes` likewise omits it, so the "bounded + observable" contract is off for every in-flight window.

**5. `ack()` clears `in_flight` before deleting the file — `src/spool.rs:179`**
`self.in_flight.take()` runs first (→ `None`), then `fs::remove_file` can fail (EIO/EACCES) and propagate via `?`. A caller that retries `ack()` gets `Ok(false)` ("nothing in flight") and moves on; `spool-N.inflight` lingers uncounted until the next `open()` re-delivers it (duplicate), and the next `drain_oldest` renames another segment to `.inflight`, leaving two at once. Delete first, clear `in_flight` only on success.

**6. Dead branch in `enforce_cap` that fails open — `src/spool.rs:232`**
`head_seq` is always the highest segment number, so `oldest == self.head_seq && seqs.len() > 1` is unreachable today. The branch `return Ok(())` (cap not enforced) rather than fail-safe means a future change to seq allocation silently disables shedding. Remove it or make it an assertion.

**7. Doc says a second drain "blocks until the first is ack'd" — `src/spool.rs:42`**
It doesn't block; it re-delivers the same segment. The behavior is fine, the word is misleading.

PR #158, opened 2026-09-10
feat(transport): implement mTLS client and event upload

### LINKED ISSUE #24: Build transport: mTLS agent-server channel

SherlockOmss:
## Summary

- **TransportClient**: HTTP client with mTLS support using ureq v3
- **EventUploader<D>**: Generic uploader with `EventDrain` trait for spool decoupling
- **TransportConfig**: mTLS paths, endpoints, retry settings
- **TransportError**: Error types with `is_retryable()` classification

## Design Decisions

1. **EventDrain trait**: Transport depends only on `schema` (leaf crate rule). Agent wires `EventSpool` via trait.
2. **ureq v3**: Built-in TLS abstraction for mTLS
3. **Exponential backoff**: 1s base, 60s max, doubles on each failure

## Test plan

- [x] `cargo check -p transport` passes
- [x] `cargo clippy -p transport -- -D warnings` passes  
- [x] `python3 tools/check-deps.py` - 38 crates OK
- [x] Unit tests pass (3 tests)
- [x] `cargo check --workspace` passes

Closes #24


old-dov:
Review pass (level `high`). Solid start on the crate skeleton and the `EventDrain` seam keeps the dependency direction clean. A few blockers before this can be wired in, mostly around the store-and-forward contract that #24 exists to guarantee.

## Blocking

**1. `upload_once` permanently loses drained events on any upload failure — `src/upload.rs:68`**
`store::EventSpool::drain_oldest` deletes the segment file before it returns (`store/src/spool.rs:88-92`). `upload_once` drains everything, then on a transient network error / 4xx / unparseable response returns `Err` without re-spooling or acking — those events are gone. With 250 events and `batch_size` 100, if batch 2 fails after batch 1 succeeded, batches 2–3 (150 events) are dropped while the caller retries with a fresh drain. This is exactly the failure mode #24 ("store-and-forward … degrade gracefully when the server is unreachable") is meant to prevent. This ties into the two-phase drain/ack work in #156 — see point 6.

**2. mTLS client certificate chain truncated to the leaf — `src/client.rs:131`**
`ureq::tls::Certificate::from_pem` is documented as "parse_pem followed by picking the first certificate". A client cert PEM that bundles leaf + intermediate (the common case for an internal PKI) will present only the leaf; server-side chain verification fails and the agent can't authenticate. `ClientCert::new_with_certs` takes a slice specifically to carry the chain — use `parse_pem` and collect all certificates.

**3. Empty / non-JSON 2xx body misclassified as a retryable network error — `src/client.rs:79`**
`post_json` always calls `response.into_body().read_json()`. `send_heartbeat` asks for a `serde_json::Value`; heartbeat endpoints commonly answer `200`/`204` with an empty body. `serde_json` errors on empty input ("EOF while parsing a value"), which maps to `TransportError::Network` → `is_retryable() == true`. Every heartbeat is then treated as a failure (spurious warnings + backoff). `upload_events` has the same issue if the server acks a batch with an empty `200`.

**4. `UploadLoop` can never be stopped — `src/upload.rs:205`**
`run(&mut self)` enters `while self.running` with a blocking `thread::sleep` (up to 60s). `stop(&mut self)` also needs `&mut self`, so it can't be called from another thread while `run()` executes, and `running: bool` is plain. `stop()` / `is_running()` are dead API and agent shutdown will hang. Needs a shared `Arc<AtomicBool>`.

**5. `is_retryable()` returns false for 429 / 408 — `src/error.rs:42`**
`ServerError` is retryable only for `status >= 500`. A rate-limited ingest endpoint returns `429`; `is_retryable()` says false → `upload_once` returns a non-retryable error → combined with point 1, the drained batch is dropped exactly when the server is asking the agent to slow down, not stop.

## Non-blocking

**6. Module docs describe a drain/ack protocol that doesn't exist — `src/lib.rs:62`**
The "At-least-once Delivery" section claims `store::EventSpool` exposes `drain_oldest()` + `ack()` with delete-after-success and crash re-delivery. That API is what #156 adds; the `EventDrain` trait in this PR only has `drain()`. Worth reconciling with #156 so the two PRs agree on the contract (and the doc matches the code that ships).

**7. `deny.toml` not updated for the new TLS stack — `Cargo.toml:18`**
`ureq` 3 → `rustls` → `ring` 0.17 and `webpki-roots` 1.0. Neither is in the `[licenses]` allow-list or `[licenses.exceptions]` (the existing exception is for `webpki-root-certs`, a different crate). `cargo deny check licenses` will flag these. The `deny.toml:43` comment ("Pulled only at build time by ort's downloader") also stops being true once `ureq → rustls` is a shipped agent dependency.

**8. `ureq` pulled with default features — `Cargo.toml:18`**
`features = ["json"]` keeps cookie handling (`cookie`, `cookie_store`, `time`), gzip (`flate2`, `miniz_oxide`), `charset`, full `idna`/`url` — ~45 transitive crates the agent doesn't use. `default-features = false` plus `json` and the rustls/tls feature trims the supply-chain surface for an on-device agent.

**9. `ingest_url()` / `heartbeat_url()` double-slash on a trailing-slash `server_url` — `src/config.rs:88`**
`TransportConfig::new("https://api.example.com/")` → `"https://api.example.com//api/v1/ingest/events"`. Many routers / strict proxies 404 or redirect on the empty path segment. Trim a trailing `/`.

**10. `EventUploader::new` missing `#[must_use]` — `src/upload.rs:43`**
`UploadLoop::new` and `TransportConfig::new` have it; `must_use_candidate` is an enforced warn and clippy runs `-D warnings`, so `cargo clippy -p transport -- -D warnings` fails.

PR #160, opened 2026-09-10
chore: run rustfmt on nightly + apply the canonical formatting

old-dov:
`rustfmt.toml` sets `group_imports` and `imports_granularity` — both nightly-only. The `lint` job ran `dtolnay/rust-toolchain@stable` + `cargo fmt --all --check`, which **silently ignores** unstable options, so the canonical import grouping was never enforced and never fully applied. `lab/lint.sh` (which does use `cargo +nightly fmt`) has been red the whole time.

## Commits

1. **`chore(ci): run rustfmt on nightly`** — the `rustfmt` step installs nightly rustfmt and runs `cargo +nightly fmt --all --check`, matching `lab/lint.sh`. `cargo doc` / `check-deps.py` stay on stable.

2. **`gitattributes: pin .rs to LF`** — `crates/correlator/src/engine.rs` and `crates/policy/src/lib.rs` were committed with CRLF while the other 104 `.rs` files are LF. rustfmt on Linux/CI rewrites those two whole just to flip line endings. New `.gitattributes` with `*.rs text eol=lf` + renormalize fixes it (this repo had no `.gitattributes` before).

3. **`chore: apply cargo +nightly fmt --all`** — 22 files, 113/78. Import regrouping (std / external / crate) and crate-granularity merging (`use a::b; use a::c;` → `use a::{b, c};`), plus a few line-wrap adjustments in `engine.rs` and `ebpf/src/main.rs`. **No semantic changes.**

## Verification

`cargo +nightly fmt --all --check` clean. Formatting can't change compilation; `cargo build`/`test` unaffected.

## Heads-up for open PRs

#155 / #156 / #157 / #158 touch `.rs` files this reformats — rebase onto `main` after this lands and re-run `cargo +nightly fmt` (or `lab/lint.sh`). The conflicts are formatting-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #161, opened 2026-09-10
rules: gate SELF-SPAWN (T1059) to Windows events (#159)

### LINKED ISSUE #159: rules: SELF-SPAWN (T1059) over-fires on benign Linux shell loops — no Linux calibration
Split off from the PR #155 hardware validation (kernels 5.15 / 6.1). Not a regression — the SELF-SPAWN rule predates #155.

## What fires

`crates/rules/src/state.rs` `check_self_spawn` (T1059): `SELF_SPAWN_THRESHOLD = 3` execs of the same `comm` from the same `ppid` within `SELF_SPAWN_WINDOW_NS = 30s` (`crates/rules/src/exclusions.rs`).

Running `lab/scenarios/argv.sh` or `lab/scenarios/lineage.sh`, the rule alerts on the scenario's own shell loop:

```
T1059 — pid=<sh> comm=sh spawned 3x in 30s by ppid=<loop> — suspected self-spawn
```

`argv.sh` is a `for i in $(seq 1 3); do sh -c "..."; sleep 1; done` — three `sh` execs in ~3s under one parent, which is exactly threshold-in-window. `lineage.sh` does the same for `nginx` / `sh` / `sleep`.

`ppid` is correct in every case (the scenario's `bash` loop), so #53 / #111 lineage is intact — this is purely the self-spawn heuristic being over-eager.

## Root cause

`check_self_spawn` sits under the `// ── Windows rules ──` section and every entry in `SELF_SPAWN_EXCLUSIONS` / `SELF_SPAWN_PARENT_EXCLUSIONS` is a Windows `.exe` (MpCmdRun, WerFault, RuntimeBroker, …). But `RuleState::on_exec` runs it unconditionally for every `ExecEvent`, Linux included, with no Linux-side calibration. On Linux a tight benign shell loop (`while true; do sh -c …; done`, a `make`/CI step, a health-check cron firing fast) trips `sh`/`bash` x3 in 30s trivially.

Same false-positive class already catalogued for Windows in `exclusions.rs` (WER respawn loop, RuntimeBroker → PowerShell).

## Direction (not prescriptive)

Whatever lands must stay evidence-gated (`policy::name_exclusion_applies`), bounded, and sliding-window — no name-only bypass. Options to weigh:

- Require the repeated child to be **short-lived AND re-spawned after the previous one exited** (an actual respawn signature), not merely N execs in a window — a `for` loop that spawns sequentially looks different from a supervisor restarting a crashing implant only in liveness.
- Gate on the parent **not** being an interactive / loop shell whose own parent is a terminal or a known build tool.
- Linux-specific threshold/window (higher count, or require the child `comm == parent comm` for true self-spawn vs. parent-spawn).
- A `SELF_SPAWN_PARENT_EXCLUSIONS` equivalent for Linux build/CI tooling (`make`, `sh`, `bash` when parented by a TTY).

## Done when

- A benign shell `while`/`for` loop that spawns the same short-lived child a handful of times in <30s no longer alerts.
- An actual re-spawning implant (see `lab/scenarios/respawn-beacon.sh`) still alerts.
- Regression test named after the behavior (per `docs/development/code-style.md`), e.g. `self_spawn_ignores_sequential_shell_loop` alongside the existing `self_spawn_window_slides_instead_of_resetting`.
- `lab/validate-155.sh` "known non-blocking noise" note and the `argv.sh` / `lineage.sh` expectations updated once it's quiet.

old-dov:
Closes #159. Split off from the #155 hardware validation, which called out a `T1059 … suspected self-spawn` firing on the scenario shell loops as "known non-blocking noise".

## The false positive

`RuleState::check_self_spawn` (T1059, 3 spawns of the same `comm` from one `ppid` in 30s) sits under `// ── Windows rules ──`: every entry in `SELF_SPAWN_EXCLUSIONS` / `SELF_SPAWN_PARENT_EXCLUSIONS` is a Windows `.exe` dated to a Windows lab capture, and unlike its siblings (`check_parent_suspect`, `check_lolbin`) it has **no comm allowlist**. But `on_exec` ran it for every `ExecEvent`, Linux included — so any script that re-spawns the same helper a few times in a loop trips it:

```
for i in $(seq 1 3); do sh -c "…"; done      # 3 execs of `sh` from one ppid, <4s
```

`ppid` is correct, so this isn't a lineage bug — it's the heuristic being over-eager with no Linux calibration.

## The fix

Early return unless `event.meta.user` is `User::Windows`. A Linux respawn that matters still surfaces through the other rules that gate on the actual malicious behaviour rather than raw spawn count:

- web-shell lineage — `check_web_server_spawns_shell` (T1059)
- download → exec — `check_download_then_exec` (T1105)
- the correlator's `rule_respawn_connect` (spawn count **+** network, keyed on `(ppid, comm)`)

A Linux-calibrated SELF-SPAWN (respawn signature: previous instance dead, irregular cadence, non-shell parent) is a separate pass if wanted.

## Changes

- `state.rs` — the gate + doc explaining why.
- `tests/mod.rs` — `exec_event_win` helper (`User::Windows`).
- `tests/windows.rs` — the eight SELF-SPAWN cases build `User::Windows` events; new `self_spawn_stays_quiet_on_a_unix_shell_loop` (5 spawns of `sh` from one ppid, no `T1059`).
- `lab/scenarios/respawn-beacon.sh` — note the rule is Windows-only now.

## Verification

`cargo +stable test -p rules` — 37 pass. `cargo +stable clippy -p rules --all-targets -- -D warnings` — clean.

## Notes

- Branched from `main`. Trivial conflict with #160 on the `use schema::…` import line (both touch it); rebase + `cargo +nightly fmt` after either lands.
- `lab/validate-155.sh`'s "known non-blocking noise" echo is on the #155 branch, not here — drop it once both #155 and this land.


Sollykhan (approved):
Clear-cut fix — the SELF_SPAWN_EXCLUSIONS / SELF_SPAWN_PARENT_EXCLUSIONS lists were already de facto Windows-only, so this closes a latent bug rather than narrowing behavior. The self_spawn_stays_quiet_on_a_unix_shell_loop test locks it in properly. Linux respawn coverage via check_web_server_spawns_shell + check_download_then_exec + correlator's 
ule_respawn_connect is a reasonable interim; a proper Linux-calibrated SELF-SPAWN can come later if the signal justifies it. Rebase after #160 for the use schema::… fmt conflict.

PR #171, opened 2026-09-11
fix(enrich,tamper): sha2 0.11 dropped impl Write for Sha256, breaks main

old-dov:
## Problem

`main` currently fails to build `crates/enrich` and `crates/tamper` (and transitively `agent`, which depends on `enrich`):

```
error[E0277]: the trait bound `Sha256: std::io::Write` is not satisfied
```

Bisected to #163 (`sha2` 0.10.9 → 0.11.0, merged this morning) — that release dropped `impl Write for Sha256`. Both crates' `hash_file` helpers streamed a file into the hasher via `std::io::copy(&mut reader, &mut hasher)`, which needs that impl.

## Fix

Read into a stack buffer and feed `Digest::update` directly instead of going through `io::copy`. Same streaming behavior — `tamper`'s `MAX_VERIFY_BYTES` cap via `Read::take` is untouched, `enrich`'s caller-side `MAX_HASH_BYTES` check is untouched. No new dependency.

## Verified

- `cargo test -p enrich -p tamper -p agent` — all green (`enrich`'s `hashes_known_vector` confirms the digest itself is unchanged, not just "compiles")
- `cargo +nightly fmt --check` and `cargo clippy --all-targets -- -D warnings` — clean on both crates

Left `Cargo.lock` untouched — running cargo locally normalized some unrelated `windows-*` duplicate-version entries, not worth carrying in this diff.


SherlockOmss (approved):
**APPROVED** ✅ Critical fix for sha2 0.11 breaking change

## Review Summary

This PR correctly addresses the breaking change in sha2 0.11 where `impl Write for Sha256` was removed.

## Code Quality
✅ Compilation successful  
✅ Proper replacement: `io::copy()` → manual `read()` + `update()` loop  
✅ Buffer size (64KB) is reasonable for file hashing  
✅ Error handling preserved with `.ok()?`  
✅ Clear inline comments explaining the breaking change  
✅ Consistent implementation in both `enrich` and `tamper` crates  

## Dependencies
✅ Both crates correctly specify sha2 = "0.11"  
✅ No transitive dependency conflicts  

## Merge Decision
**Ready to merge immediately** - this fixes a build breakage on main.

Issue #93, opened 2026-09-03
Build sensor-linux-journal: journald auth + service lifecycle


florianamette:
**Component:** `crates/sensors/linux/journal` · see crate doc.

sd-journal tail on an allowlist: sshd accept/fail, PAM sessions, sudo/su, unit lifecycle. Normalized with provenance; schema additions additive (auth/session event type shared with Windows 4624 and macOS login work).

## Done when
- [x] SSH login + failed-login burst on a lab VM land as normalized events
- [x] A service install/start is case-attachable context

old-dov:
PR #238 continues this foundation: sshd accept/fail, sudo, and su now map into the shared `schema::AuthEvent` (issue #94/ADR-0005 landed since this foundation was written, unblocking the cross-platform decision this PR's original doc was waiting on), wired into `agent run` alongside the netlink poller. Validated on real hardware (Hyper-V lab, Arch) — SSH login, a 3-attempt failed-login burst, and sudo all land as normalized events with correct attribution. Also found a real classifier bug in the foundation while validating: this box's OpenSSH logs the per-connection worker as `sshd-session`, not `sshd` — fixed in the PR, full writeup there.

First `Done when` item closed. Second (service install/start as case-attachable context) deliberately left open — see the PR's "out of scope" section, it needs the same kind of cross-platform schema decision `AuthEvent` did, and Nikolas is already working the Windows side of that exact problem in #224.

old-dov:
Second checkbox closed by #287 (`UnitPersistenceTracker` + T1543.002 rule, merged into main at 397f383). Closing.

Issue #91, opened 2026-09-03
Build sensor-linux-lsm: BPF-LSM observation + inline blocking


florianamette:
**Component:** `crates/sensors/linux/lsm` · see crate doc. The io_uring answer AND the Linux prevention path.

LSM-hook programs (bprm_check, file_open, socket_connect, ...) observing at the security layer regardless of entry path; -EPERM capability wired to `response` verdicts (policy-gated, default observe). CONFIG_BPF_LSM detected and reported honestly via capabilities/conformance.

## Done when
- [x] Lab: an io_uring-submitted file write is observed (tracepoint-only build misses it — the regression proving the point)
- [ ] A policy-enabled block denies a test exec with a detection recording the denial
- [ ] Capability reporting correct on kernels without lsm=bpf

old-dov:
PR #239 closes the first `Done when` item on real hardware (Hyper-V lab, Arch `7.2.4-arch1-2`): the `file_open` LSM hook fires for a real `fio --ioengine=io_uring` write. Found and fixed a real verifier bug in the already-merged program along the way — full writeup in the PR and the crate doc. Second and third items stay open, blocked for the reasons already documented in the crate doc (verdict model doesn't exist yet for #2; no schema surface + no production caller for #3) — not attempted, not guessed at.

PR #173, opened 2026-09-11
sensor-linux-lsm: file_open LSM hook, observation only (issue #91, foundation)

old-dov:
## Scope

First slice of #91 (BPF-LSM observation + inline blocking). Scoped down hard after checking what's actually there to build on — see the commit message for the full reasoning. Short version: `crates/response`/`crates/policy` have no verdict model yet (#131, #133), so there's nothing for `-EPERM` to consult, and this box can't validate a real attach (no root here, and no `/sys/kernel/security/lsm` to confirm "bpf" is an active LSM even though the BTF type for the hook is compiled in — checked empirically).

## What's here

- **`crates/sensors/linux/ebpf`**: `file_open` LSM hook, compiled into the same object as the tracepoint probes. Always returns 0 (or defers to an earlier LSM program's verdict) — never denies. Counts hits in a new per-CPU `LSM_FILE_OPEN_HITS` map. No `vmlinux` bindings needed — reads no `struct file` field, same discipline as the rest of the probe.
- **`crates/sensors/linux/lsm`** (was an empty stub): `detect_hook_support` — root-free, no `bpf(2)` syscall, BTF-only capability check (necessary, not sufficient — documented why). `attach_file_open`/`file_open_hit_count` — load/attach/read against an already-loaded `Ebpf` (reuses `sensor_linux::load_ebpf()`).
- `docs/sensors/linux.md` updated.

## Deliberately not here (follow-ups)

- Inline blocking (needs `response`/`policy` verdict plumbing that doesn't exist)
- `bprm_check`/`socket_connect` hooks
- A real event feed from this path — today's counter is a liveness signal, not telemetry; an LSM-sourced open and the existing tracepoint-sourced `FileOpenEvent` need a decided dedup relationship first
- All three of #91's lab-validated "Done when" items (io_uring write observed, policy-gated block, capability reporting on a kernel without `lsm=bpf`) — need the Hyper-V lab kernels

## Verified

- `cargo +nightly fmt --check` and `cargo clippy --all-targets -- -D warnings` — clean on `sensor-linux`/`sensor-linux-lsm`
- `cargo test` — sensor-linux 17/17 unaffected; `sensor-linux-lsm`'s one test is deliberately portable (asserts a nonexistent hook name is rejected, not a specific outcome for `file_open` — that depends on the test machine's kernel config)
- The eBPF cross-compile to `bpfel-unknown-none` accepts the new LSM program

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #175, opened 2026-09-11
sensor-linux-journal: journald allowlist tail, no shared event type yet (issue #93, foundation)

old-dov:
## Scope

First slice of #93. Scoped down after checking what's there to build on: the issue's own text says the auth/session event type is "shared with Windows 4624 and macOS login work" — but `schema` has no auth/login event at all today (checked). Designing one is a cross-platform call, not something to bake into a Linux-only PR unreviewed. Same discipline as #91's foundation.

## What's here

- **`crates/sensors/linux/journal::record`** — parses `journalctl -o json` lines field-by-field (not a derived struct), so a binary-safe `MESSAGE` (journald's byte-array form for non-UTF8 content) is skipped, not a hard parse failure.
- **`crates/sensors/linux/journal::classify`** — allowlist: sshd accept/fail, PAM session open/close (any `pam_unix` service, not sudo-specific), sudo command lines, systemd unit start/stop/fail. Unit lifecycle keys off the readable `JOB_TYPE`/`JOB_RESULT` fields rather than the opaque systemd message-catalog UUID — I could verify the former against a real capture, and it's auditable without looking up a hex constant.
- **`crates/sensors/linux/journal::tail`** — `ClassifiedJournal`, a `BufRead`-generic iterator (testable without real journald), plus Linux-only `spawn_follow`/`current_cursor` driving real `journalctl -f -o json --after-cursor=...`. Subprocess, not `libsystemd` FFI — no build-time `libsystemd-dev` dependency, same spirit as #152 dropping the vmlinux bindings crate for `/proc` reads.
- `docs/sensors/linux.md` updated.

## What's verified vs. documented-but-unverified

Real captures from this dev machine's own journal: the sudo command-line fixture, the PAM session open/close fixtures, and the unit-lifecycle `JOB_TYPE`/`JOB_RESULT`/`USER_UNIT` shape. sshd accept/fail and `su` are built against the standard, documented OpenSSH/PAM log formats but **not** lab-verified — no `sshd` on this box, `su` needs interactive auth this session can't provide. Called out explicitly rather than assumed correct.

## Deliberately not here (follow-ups)

- A `schema::Event` variant (needs the cross-platform design pass above)
- A `Sensor` trait implementation (nothing to push without an event type)
- Cursor persistence across restarts (`crates/store` integration)
- Lab validation of the sshd/su paths

## Verified

- `cargo +nightly fmt --check` and `cargo clippy --all-targets -- -D warnings` — clean on `sensor-linux-journal`
- `cargo test --locked` — 26/26 green, including a real `journalctl -f` + `current_cursor` integration test that skips itself if `journalctl` isn't on `PATH` rather than asserting a specific journal history
- Full-workspace `cargo test` is blocked in this sandbox by a missing `pkg-config` (an unrelated `openssl-sys` build dependency, pre-existing, not touched by this PR)


SherlockOmss (approved):
**APPROVED** ✅ Solid journald foundation, disciplined scope

## Review Summary
First slice of issue #93 - journald ingestion foundation. Excellent scope discipline: deliberately defers auth/session event type design until cross-platform consensus, same as #91 and #173.

## Code Quality  
✅ **Compilation**: Clean build, clippy passes  
✅ **Architecture**: Subprocess `journalctl` instead of libsystemd FFI (no build-time deps, matches #152 spirit)  
✅ **Parsing**: Field-by-field `JournalRecord` (binary-safe, non-UTF8 `MESSAGE` skipped gracefully)  
✅ **Classification**: Allowlist-based (sshd, PAM, sudo, systemd unit lifecycle)  
✅ **Integration test**: `spawn_follow` + `current_cursor` validated, self-skips if no `journalctl`  

## Design Discipline  
✅ **Cross-platform deferred**: No auth/session Event type yet (Windows 4624 / macOS login need joint design)  
✅ **No premature Sensor impl**: Can't push events that don't exist  
✅ **Cursor persistence deferred**: Needs `store` integration (follow-up)  
✅ **Documented verification gaps**: sshd/su paths not lab-tested (no sshd, su needs interactive auth)  

## Technical Correctness  
✅ Parses `journalctl -o json` (subprocess stdout)  
✅ Unit lifecycle keys off readable `JOB_TYPE`/`JOB_RESULT` (not opaque UUIDs)  
✅ Real captures verified (sudo, PAM session from dev machine journal)  
✅ Standard-based but unverified (sshd/su - documented as such)  

## Testing  
✅ 26 tests (PR description claims - cannot run without sudo)  
✅ Real journalctl integration test with conditional skip  

## Dependencies  
- Depends on: nothing (infrastructure only, no Event variant yet)  
- Blocks: #93 follow-up slices (Event design, Sensor impl, cursor persistence)  

## Merge Decision  
**Ready to merge** - foundation is solid, scope discipline exemplary, follows #91/#173 pattern. Auth event type design is correctly deferred to cross-platform thread.

PR #176, opened 2026-09-11
sensor-linux-netlink: sock_diag foundation, conntrack/proc-connector deferred (issue #92)

old-dov:
## Scope

First slice of #92. Only `sock_diag` of the issue's three netlink sub-protocols — it's the one confirmed reachable unprivileged (verified empirically: TCP `sock_diag` works without root here, matching `ss`), and its wire format is a fixed-size struct rather than TLV-nested attributes. conntrack and proc connector are both technically reachable in this sandbox (this session has passwordless `sudo`) but each is a distinct sub-protocol with its own parser — scoped out to keep this reviewable, not silently dropped.

## What's here

- **`crates/sensors/linux/netlink::wire`** — pure encode/decode for `inet_diag_req_v2`/`inet_diag_msg`. No `#[repr(C)]` transmute; every field offset — including the easy-to-miss detail that `idiag_sport`/`idiag_dport` are big-endian while the rest of the struct is host-order — is verified against a real captured response from this dev machine's own kernel.
- **`crates/sensors/linux/netlink::socket`** (Linux-only) — opens the raw `NETLINK_SOCK_DIAG` socket, sends the dump request, loops `recv` until `NLMSG_DONE`. The loop is load-bearing: a single `recv()` empirically did **not** return the terminator in testing, only 4 of the records — multi-call draining is a correctness requirement here, not defensive padding.
- **`crates/sensors/linux/netlink::proc_join`** — inode→PID(s) join via `/proc/<pid>/fd/*` → `socket:[<inode>]`, same technique `ss`/`lsof` use. A multi-map on purpose (shared listeners across forked workers) and explicitly partial without root (another user's `/proc/<pid>/fd` is `EACCES`, treated as "no PID found," not an error).
- `snapshot()` at the crate root joins the two.
- `docs/sensors/linux.md` updated.

## Deliberately not here (follow-ups on #92)

- conntrack (nfnetlink, TLV-nested tuple/counter attributes)
- proc connector (fork/exec/exit — the eBPF cross-check)
- UDP sockets ("listen/established" is TCP-state terminology per the issue's own wording)
- Periodic re-snapshotting / drift detection (no lab VM here to validate against)
- `schema::Event`/`Sensor` wiring — beacon volume/periodicity features and the eBPF cross-check both need `correlator`/`tamper` plumbing that doesn't exist yet

## Verified

- `cargo +nightly fmt --check` and `cargo clippy --all-targets -- -D warnings` — clean
[...]

Issue #92, opened 2026-09-03
Build sensor-linux-netlink: sock_diag, conntrack, proc connector


florianamette:
**Component:** `crates/sensors/linux/netlink` · see crate doc.

Periodic sock_diag snapshots (listen/established + inode→pid), conntrack flow accounting (bytes/duration per flow → beacon volume/periodicity features), proc-connector cross-check against the eBPF stream (divergence = tamper signal for `tamper`).

## Done when
- [x] Listening-port drift on a lab VM produces events
- [x] Conntrack features reach the correlator for a beacon scenario
- [x] Cross-check divergence test (suppress eBPF, watch the signal)

old-dov:
Tried to validate the two open \`Done when\` items (conntrack features reaching a beacon detection, cross-check divergence with eBPF suppressed) today, in WSL2 (kernel 6.6.114-microsoft-standard-WSL2). Listen-port drift (item 1) fires reliably — saw \`T1571\` multiple times, no issue there.

The beacon-via-conntrack path did not fire in several attempts, and digging into why turned up a broader problem with the environment rather than the code:

- \`ss -tn state established\` reports **nothing** for connections independently confirmed \`ESTABLISHED\`/present in the kernel's own \`/proc/net/nf_conntrack\` table at the same moment. \`attribute_flow\` (\`crates/sensors/linux/netlink/src/normalize.rs\`) requires a socket-diag entry with \`SocketState::Established\` to join a conntrack flow to a pid — if sock_diag itself doesn't surface the socket, the join can never succeed, regardless of poll timing.
- This is on top of the already-known WSL2 defect where every process reports \`ppid=0\` (see the PR #229 thread).

I don't think this says anything about the real code — WSL2 isn't a \`lab/MATRIX.md\` row, and I couldn't get a clean read even after correcting my own test-script mistakes along the way (premature connection closes, an orphaned test listener eating CPU, quoting bugs). I'm stopping here rather than chase it further in this environment. Worth re-running the beacon-via-conntrack case in the Hyper-V lab (5.15/6.1, where \`ss\`/sock_diag are known-good from the #53/#152 validations) before trusting either a pass or a fail from WSL2 on this one.

old-dov:
[...]

PR #179, opened 2026-09-14
sensor-linux-netlink: proc connector foundation (issue #92)

old-dov:
Part of #92 (netlink sensor). Second foundation slice, sibling of #176's sock_diag work — same "reviewable, self-contained" scope discipline.

## What lands here

`ProcEventSubscription` — a live `NETLINK_CONNECTOR` subscription to the `CN_IDX_PROC` multicast group, decoding kernel fork/exec/exit broadcasts:

- **`proc_events.rs`** — pure wire-level encode/decode for `cn_msg` + `struct proc_event`. Allocation-light, unit-tested against synthetic byte layouts pinned to `<linux/connector.h>`/`<linux/cn_proc.h>`.
- **`proc_socket.rs`** — drives the real `AF_NETLINK`/`NETLINK_CONNECTOR` socket: open, bind to the process-events group, send the `PROC_CN_MCAST_LISTEN` subscribe, `recv_events()` in a loop. Root/`CAP_NET_ADMIN`-only (unlike `sock_diag`): confirmed against this dev machine that the kernel rejects an unprivileged subscribe with `EPERM`, surfaced as a normal `NetlinkError`, not a panic.

## Verification

Full suite as root (WSL2, `edr-lab`): **23/23 tests green**, `cargo clippy --all-targets -- -D warnings` clean, `cargo +nightly fmt --check` clean.

The live integration test (`observes_live_fork_exec_and_exit_broadcasts`) does **not** assert against one specific spawned child's pid — deliberately, same discipline as `socket::query_tcp_sockets_against_the_real_kernel` ("doesn't assert a specific socket is present"). While scoping this I confirmed via an ad hoc reference client that a dev sandbox under concurrent load (this machine: dozens of unrelated fork/exec/exit broadcasts/sec) can miss one specific short-lived child's own broadcasts to a freshly-opened subscription — a real race in a busy environment, not a decode bug. What the test proves is the live plumbing (socket → subscribe → recv → decode) end to end against the real kernel; a spawned child just guarantees some fresh activity exists to observe within the window.

Non-root path skips gracefully (same discipline as `sensor-linux-journal`'s `journalctl`-presence skip, issue #93's precedent) rather than requiring a specific dev/CI privilege level to build and test.

**A real bug the truncated-body test caught before it shipped:** the initial `match` used length guards (`PROC_EVENT_FORK if body.len() >= 16`), which silently fell through to the `Other(_)` catch-all when a *recognized* `what` had a too-short body — decoding a truncated `FORK` as an opaque `Other(PROC_EVENT_FORK)` instead of rejecting it. Fixed: each arm checks its own length and returns `None` explicitly.

## What is *not* touched

- **conntrack** — its own, meaningfully bigger netlink sub-protocol (TLV-nested attributes: tuples, counters), scoped out to keep this slice reviewable. Tracked as a separate #92 follow-up, not silently dropped.
- No `schema::Event` variant or `schema::sensor::Sensor` implementation — the eBPF cross-check and beacon volume/periodicity features both need `crates/correlator`/`crates/tamper` wiring that doesn't exist yet.
- `PROC_EVENT_UID`/`_GID`/`_SID`/`_PTRACE`/`_COMM`/`_COREDUMP` — decoded as `ProcEvent::Other` rather than their own variants; nothing in the eBPF stream to cross-check them against yet.
- Periodic drift detection / a lab-VM "listening-port drift" validation — out of scope for this slice, same as #176.


SherlockOmss (approved):
## Review: APPROVED

Comprehensive review completed. This PR delivers solid foundation work for proc connector integration.

### Key Strengths

**Code Quality:**
- Wire protocol parsing is meticulous and correct (cn_msg + proc_event layouts match kernel headers)
- Excellent bug fix: truncated event bodies now properly rejected vs. silently misclassified
- All 13 unsafe blocks have comprehensive SAFETY comments
- Memory-safe: proper bounds checking, OwnedFd for resource management
- Error handling: proper thiserror usage, no unwrap/expect in production paths

**Testing:**
- 8 unit tests cover all event types (FORK/EXEC/EXIT) + truncation detection
- Integration test gracefully skips on EPERM (non-root)
- Test design is sound for busy/concurrent environments

**Project Compliance:**
- Dependency rules: PASS (sensor depends only on libc)
- Platform gating: correct cfg(target_os = "linux")
- Unsafe documentation: excellent

### Minor Notes (non-blocking)

- Some pub struct fields lack individual doc comments (consistent with existing codebase patterns)
- Cannot verify test execution (sudo required), but test design is sound

### Scope Discipline

Correctly defers conntrack (bigger TLV parser) and schema::Event integration (needs correlator/tamper wiring). Foundation slice is reviewable and self-contained as intended.

**Recommendation:** Merge. Ready for production.

Reviewed-by: Claude Sonnet 4.5 via agent a08ee4e

PR #174, opened 2026-09-11
ml/training: trainers take --dataset, verify manifest, emit training.json (part of #44)

Sollykhan:
Part of #44 (ML data infrastructure). Third slice, stacked on top of #172 (training record) which is stacked on top of #170 (manifest sidecar). Merge in order.

## Where we were

`train_linux.py` and `train_windows.py` used to read a hard-coded path (`training/data/baseline_benign.jsonl` and friends), fit an Isolation Forest, and write `model.onnx` next to themselves. The existing registry cards for those runs admit it out loud in the card body:

> Training data: Linux/Windows lab baseline capture (old iteration; dataset not versioned under the current datasets/ scheme - **reproducibility criterion NOT met**).

There was no mechanical link from a shipped model back to the baseline it was trained on.

## What lands here

Both trainers now take:

```
--dataset <baseline_dir> [<baseline_dir> ...]   # one or more (nargs="+")
--output-dir <registry_version_dir>
```

and, on every run:

1. **Verify each manifest before touching anything else** (via `dataset_version_from_manifest` from #172) - a mutated baseline aborts the run before a partial `model.onnx` is written to disk.
2. **Concatenate + re-dedupe** samples across the supplied datasets so a command line captured on two machines does not double-weight the boundary.
3. **Write `model.onnx` AND `training.json`** to `--output-dir`, with `dataset_versions` populated straight from the verified manifests.

Rule 3 of `ml/README.md` is now mechanically enforced end to end: manifest sidecar (#170) -> DatasetVersion binding (#172) -> trainer emits `training.json` naming them.

Multi-dataset support (`nargs="+"`) is what the Windows benign campaign is going to need: capture on N machines, one baseline directory per machine, one `train_windows.py` run consuming them all.

### What is *not* touched

- `SANITY_CHECK_SAMPLES` on both sides is preserved as-is (Windows keys, Linux keys). `verify_onnx.py` imports the Windows one and its shape is unchanged.
- The existing `card.md` files in `ml/registry/*/0.1.0/` stay flagged "reference-only, NOT shippable" - the successor runs that swap them out use this new path.
- Feature extraction (`extract_features`), the ONNX target opset pin, and the `contamination=0.05` decision comment are all preserved; they carry incident context that new PRs should not silently drop.

## Deliverables

- `ml/synthaea_ml/training/train_linux.py`, `train_windows.py` — refactored to the new `--dataset`/`--output-dir` API.
[...]

PR #170, opened 2026-09-11
ml/data: add baseline dataset manifest sidecar (part of #44)

Sollykhan:
Part of #44 (ML data infrastructure). First slice: the manifest sidecar. Registry binding and trainer guards follow in separate PRs.

## The gap

`ml/README.md` rule 3 says a registry model records the exact dataset versions it was trained on. Today a baseline is a raw `.jsonl` with no version, provenance, or workload label - a model card that names it can silently point at a mutated file, and there is no way to bin captures by role for the per-site adaptation loop described in `docs/detection/ml.md`.

## What lands here

A `manifest.json` sidecar written next to each `baseline.jsonl`, decoupled from the sample format (so both Linux `{"argv": [...]}` and Windows `{"cmdline": "..."}` baselines share the same manifest):

```json
{
  "schema_version": 1,
  "host_id": "a1b2c3d4e5f6",
  "platform": "windows",
  "os_version": "11 24H2",
  "workload_label": "desktop-user",
  "capture_start": "2026-09-11T10:00:00Z",
  "capture_end":   "2026-09-11T11:00:00Z",
  "sample_count": 3421,
  "sample_sha256": "…",
  "capturer_ref": "solkapc-run-1",
  "extra": {}
}
```

### Design points

- **`host_id` = `sha256(hostname)[:12]`**, not the raw hostname. Datasets shipped between team members carry only the hash. Same host is stable across captures (per-site adaptation needs that); different hosts do not collide in any fleet size we will ever have.
- **`sample_sha256` locks the dataset version.** `verify_manifest()` re-hashes the baseline before a training run; a mismatch is the "someone edited the baseline without regenerating the manifest" case, which the (upcoming) registry glue will treat as a hard fail.
- **`workload_label` is free-form** (`desktop-user`, `dev`, `admin`, `server-idle`) - lets the adaptation loop bin captures by role. Not a typed enum: adding a new role should not require a schema bump.
- **Fully decoupled from the sample format.** `train_linux.py` and `train_windows.py` are untouched - they keep reading their JSONL as they do today. Trainer guards (refuse to train against a baseline without a valid manifest) are a separate PR so this one stays reviewable.
- **UTC-normalised, `Z`-suffixed timestamps.** A capture from a machine in another timezone is stored as the same UTC instant, so cross-timezone campaign data is comparable.

### Deliverables

[...]

Issue #44, opened 2026-09-03
ML data infrastructure: scenario corpus + benign capture campaign


florianamette:
**Component:** `ml/datasets`, lab · **Depends on:** #10 (lab migration)

The prerequisite for every ML differentiator in `docs/detection/ml.md`: robustness cards, conformal FP budgets, per-site adaptation, and fleet rarity are all only as good as the data they are measured on. Today's baselines (71 Linux lines, one Windows machine) cannot support any of them.

Two deliverables, both versioned and replayable:
- **Scenario corpus**: lab attack scenarios as first-class, versioned artifacts (provisioning + execution + expected detections), replayable for evaluation — the source of known-bad samples for the adversarial harness and of labeled traces for T2.
- **Benign capture campaign**: diverse per-platform benign corpora (multiple machines, roles, workloads), curated via `capture_to_baseline`-style tooling into recorded dataset versions under `ml/datasets/`.

## Done when
- [x] Scenario format defined; existing lab scenarios (#10) converted and replayable
- [ ] Benign corpora per platform large enough that the FP gate is statistically meaningful (target documented in `ml/datasets/README.md`)
- [x] Dataset versioning in place: a registry model names the exact dataset versions it was trained on (rule 3 of `ml/README.md`)

Sollykhan:
## Scoping — where I'd start on this

### Current state (grepping the repo)
- **Scenarios**: 6 `.sh` scripts in `lab/scenarios/` + a Markdown table pairing each with its expected detections. Structure is right conceptually but not machine-parsable.
- **Baselines**: 71 Linux command lines + a single Windows machine capture under `ml/datasets/baselines/` — far below any statistically meaningful FP budget.
- **Registry**: `cmdline-iforest-linux` and `cmdline-iforest-windows` entries exist, but neither pins the dataset version(s) it was trained on (rule 3 of `ml/README.md` is not yet enforced).
- **Storage**: `ml/datasets/README.md` says data is never committed, but no concrete backend is picked yet.

### Four design decisions to close before any code lands

1. **Scenario format**. YAML/TOML sidecar next to each `.sh`, or frontmatter inside the script? Sidecar wins for tooling (schema validation, replay engine reads without executing) but doubles the file count.
[...]

PR #180, opened 2026-09-14
ml/export: add provenance release gate to verify_onnx

Sollykhan:
## Summary

Closes the loop on rule 3 of `ml/README.md` ("no model ships without a verifiable training record"): `verify_onnx` now checks the training record's provenance against the baselines on disk, not just the ONNX model's runtime properties.

## What changes

- `verify_onnx.verify_provenance(model_dir, baselines_root, *, strict)`: reads `training.json` next to `model.onnx`, calls `verify_training_record` (from #177), fails on any mismatch.
- Legacy pre-#44 entries (`cmdline-iforest-{linux,windows}/0.1.0/`) have no `training.json` — they **warn** in dev mode, **fail** under `SYNTHAEA_STRICT_PROVENANCE=1` (CI toggle).
- New tests: 16 cases in `test_verify_provenance.py` (legacy warn/fail by mode, valid record, mutated baseline caught, missing baseline dir, env var parsing).

## Why the env var toggle

The existing registry has two legacy entries that predate #170/#177. Making strict mode default would break `python -m synthaea_ml.export.verify_onnx` for anyone running locally on those. The env var lets CI enforce strict provenance (no legacy shipping to release) without breaking local dev on the interim state.

Once the two legacy entries are retrained through the new pipeline (out of scope for this PR — likely closed as part of #44), the env var can be flipped to default-strict.

## Test plan

- [x] `ruff check` clean
- [x] `pytest ml/tests/test_verify_provenance.py -v` → 16 passed
- [x] Manual: `python -m synthaea_ml.export.verify_onnx ml/registry/cmdline-iforest-linux/0.1.0/model.onnx` → warns, exits 0 ✓
- [x] Manual: `SYNTHAEA_STRICT_PROVENANCE=1 python -m synthaea_ml.export.verify_onnx ...` → fails on legacy ✓

## Related

Part of #44 (ML data infrastructure). Depends on #170 (manifest) + #177 (training record) + #174 (trainer refactor), all merged.


Sollykhan:
CI ML dispatch attempted (run 34830378064) — startup failure, org billing blocked:

> The job was not started because recent account payments have failed or your spending limit needs to be increased.

Confirms the note in `ml.yml` (`workflow_dispatch` comment). Local validation instead:

- `ruff check ml` → clean
- `pytest ml/tests/test_verify_provenance.py -v` → 16 passed
- Manual dev mode → `WARN` + exit 0 (verified on `cmdline-iforest-linux/0.1.0`)
- Manual strict mode → `FAIL` + exit 1 (verified on `cmdline-iforest-linux/0.1.0`)

Follow-up note (not in this PR): `verify_onnx` imports `SANITY_CHECK_SAMPLES` from `train_windows` regardless of the model tested. On a Linux model, all 5 Windows cmdlines score as anomalies (unrelated tokens). Should dispatch sanity samples per platform — will open a separate issue.

old-dov:
Reviewed. Solid implementation, verified it end to end:

- 16/16 new tests pass (`pytest ml/tests/test_verify_provenance.py`), `ruff check` clean.
- Traced the exception surface: `verify_training_record`/`verify_manifest` only ever raise `FileNotFoundError`/`ValueError` per their own docstrings, and `verify_provenance` catches exactly those two — no gap where a real bug could escape as an unhandled traceback instead of a clean `FAIL`.
- Manually confirmed both CLI modes against the real `cmdline-iforest-linux` legacy entry: non-strict → `WARN ...`, exit 0; `SYNTHAEA_STRICT_PROVENANCE=1` → `FAIL ...` + `1/1 model(s) failed verification`, exit 1. Matches the PR description exactly.

One thing worth flagging, not a blocker on the code itself: I don't see `SYNTHAEA_STRICT_PROVENANCE` referenced anywhere in `.github/workflows/` (checked `ml.yml`), and Actions billing is off right now anyway (per Florian, 07/09). So today this release gate is correct but dormant — nothing actually runs it in strict mode yet, it's opt-in-by-hand. Not asking you to wire CI back up (that's not this PR's problem to solve), just flagging so nobody assumes "the release gate" is already enforced somewhere it isn't. Might be worth a one-line note in the PR description or `ml/README.md` saying it's manual-only for now, so it doesn't get lost.

No changes requested from me — good to merge as is.

PR #172, opened 2026-09-11
ml/registry: add training_record for dataset version binding (part of #44)

Sollykhan:
Part of #44 (ML data infrastructure). Second slice, stacked on top of #170 (baseline manifest) - merge #170 first, or review this one against it.

## The gap

`ml/registry/README.md` says a model card records "training data (dataset versions), metrics, FP rate, ..." and `ml/README.md` rule 3 says a registry model can be rebuilt from its recorded dataset versions. Today's cards (`cmdline-iforest-linux/0.1.0/card.md`) explicitly admit `dataset not versioned under the current datasets/ scheme - reproducibility criterion NOT met`. There is no mechanical way to answer "was this model trained on the datasets its card claims?".

## What lands here

A new module `synthaea_ml.registry.training_record` writing a `training.json` sidecar next to each `card.md`:

```
ml/registry/<model>/<version>/
    model.onnx
    card.md          # for humans (existing)
    training.json    # for machines (new)
```

```json
{
  "schema_version": 1,
  "trained_at": "2026-09-11T12:00:00Z",
  "training_script": "synthaea_ml/training/train_linux.py",
  "dataset_versions": [
    {
      "name": "linux__dev__a1b2c3d4e5f6__2026-09-09",
      "baseline_sha256": "…",
      "sample_count": 71
    }
  ],
  "hyperparameters": {"n_estimators": 100, "contamination": 0.05},
  "extra": {}
}
```

### Design points

- **Baselines locked by content hash, not by path.** `DatasetVersion.baseline_sha256` is the `sample_sha256` the manifest publishes (see #170). A dataset can move on disk without invalidating the record, but a mutated baseline cannot pass `verify_training_record()` even if it keeps the same path.
- **`dataset_version_from_manifest()` verifies before returning.** A caller cannot accidentally build a training record against an edited baseline - the manifest is checked at construction time and a mismatch throws.
- **`default_dataset_name()` derives a stable, human-legible name from the manifest**: `<platform>__<workload>__<host_id>__<YYYY-MM-DD>`. Double underscores as separator so a workload like `desktop-user` does not become ambiguous.
- **`verify_training_record()` is the release gate.** For every referenced dataset it: (1) checks the baseline still exists under `baselines_root/<name>/`, (2) verifies the manifest, (3) compares the manifest's `sample_sha256` against what the training record locked. A moved or mutated dataset is a hard fail.
[...]

PR #182, opened 2026-09-14
sensor-linux-netlink: conntrack foundation (issue #92)

old-dov:
Part of #92 (netlink sensor). Third and last #92 foundation slice — **independent of #179** (proc connector): based directly on `main`, not stacked on it, since conntrack and proc connector don't depend on each other's code. This avoids the base-branch mess #172/#177 hit this morning on the ML stack — deliberately not repeating that here.

## What lands here

`dump_conntrack()` — a one-shot `NETLINK_NETFILTER`/`ctnetlink` dump of the kernel's conntrack table (IPv4 + IPv6):

- **`conntrack_attrs.rs`** — pure decode of the `nlattr` tree `ctnetlink` uses: `CTA_TUPLE_ORIG`/`CTA_TUPLE_REPLY` → `CTA_TUPLE_IP`/`CTA_TUPLE_PROTO` → addresses/protocol/ports, `CTA_COUNTERS_ORIG`/`CTA_COUNTERS_REPLY` → packets/bytes. This recursive tree (up to 3 levels deep) is exactly what made conntrack "a full netlink sub-protocol of its own" when `sock_diag`/proc connector shipped as the first two slices — a meaningfully bigger parser than either sibling's flat struct.
- **`conntrack_socket.rs`** — drives the real socket (open/bind/send/recv), same one-shot-dump shape as `sock_diag`'s `query_tcp_sockets`.

## Verification

Full suite as root (WSL2, `edr-lab`): **21/21 tests green**, `cargo clippy --all-targets -- -D warnings` clean, `cargo +nightly fmt --check` clean.

Byte layouts in the fixtures aren't invented from the header alone — I wrote a small Python `NETLINK_NETFILTER` reference client first, captured a real dump on this dev machine (`curl http://example.com` while watching the table), decoded it field by field, and only then pinned the confirmed layout into Rust tests. That process caught two real environment facts worth flagging for whoever validates this on a target host later:

1. **`CTA_COUNTERS_*` requires `net.netfilter.nf_conntrack_acct=1`** — off by default; confirmed empirically that no counters attribute appears in the dump at all until it's turned on. This module doesn't toggle it — a beacon-detection caller needing byte/packet accounting must ensure it's enabled on the target host.
2. **Conntrack entries only exist once a netfilter ruleset references conntrack.** Loading `nf_conntrack_netlink` alone left the table empty even with real outbound traffic (`curl`) until a single `iptables -A OUTPUT -m conntrack --ctstate NEW -j ACCEPT` rule was added.

Non-root path skips gracefully (same discipline as `sensor-linux-journal`'s `journalctl`-presence skip) rather than requiring a specific dev/CI privilege level — unprivileged reachability of conntrack isn't characterized here (every capture in this sandbox ran as root).

**A real bug caught by the test suite before it shipped:** the generic `nlattr` walker didn't validate the `NLA_F_NESTED` flag before recursing into a "container" attribute's value — a malformed message claiming `CTA_TUPLE_ORIG` without the nested flag would still have been parsed as a sub-attribute tree. Fixed: recursion now checks `attr.nested` first (and the flag itself went from an unused field flagged by `dead_code` to something the parser actually validates).

## What is *not* touched

- **`CTA_PROTOINFO`** (per-protocol state, e.g. TCP's own state machine) — a third level of nesting beyond what volume/periodicity beacon features need. Tracked as a further #92 follow-up, not silently dropped.
- No `schema::Event` variant or `schema::sensor::Sensor` implementation — same as #179, needs `crates/correlator`/`crates/tamper` wiring that doesn't exist yet.
- Proc connector (`NETLINK_CONNECTOR`) — separate, independent slice, see #179.


SherlockOmss (approved):
## Review: APPROVED

Comprehensive review completed (agent affa379). PR is production-ready.

### Key Strengths

**Recursive nlattr Tree Parsing (3 levels deep):**
- Correctly handles nested attribute parsing at all levels
- Level 1: `ConntrackFlow::parse` walks top-level attributes
- Level 2: `FlowTuple::parse` recurses into `CTA_TUPLE_IP`/`CTA_TUPLE_PROTO`
- Level 3: `parse_tcp_state` handles `CTA_PROTOINFO_TCP` nesting
- Generic `parse_attrs` properly validates header size, attribute length bounds, and 4-byte alignment

**NLA_F_NESTED Flag Validation (Critical Bug Fix):**
- Implemented at all recursion points (lines 159-163, 362-363, 368-372, 306)
- Prevents malformed messages from being incorrectly parsed as nested structures
- Test coverage validates flag extraction
- Excellent catch before shipping

**Memory Safety:**
- All 5 unsafe blocks have comprehensive SAFETY documentation
- All memory accesses bounds-checked before indexing
- `parse_attrs` validates offset + length <= buffer size
- IP address parsing checks exact size before conversion
- No panics on malformed input - defensive parsing throughout

**Test Coverage:**
- 10 comprehensive tests (9 in conntrack_attrs, 1 integration)
- Real capture data verified against Python reference client
- Covers full end-to-end flow, TCP state, counters, IPv6, truncation, error cases
- Integration test gracefully skips when kernel refuses request

**Documentation:**
- Excellent module-level docs explaining recursive tree structure
- Big-endian values vs host-order headers clearly documented with real example
- All public types documented
- nf_conntrack_acct=1 requirement properly flagged

### Verification

- Dependency rules: PASS (38 crates)
- Code style: All conventions followed (pub docs, SAFETY comments, no unwrap on untrusted input)

### PR Claims Verified

- Python reference client cross-checking: Documented in tests
- Bug fix NLA_F_NESTED: Properly implemented
- CTA_COUNTERS requires nf_conntrack_acct=1: Thoroughly documented
- Scope discipline: CTA_PROTOINFO deferred correctly

**This is the most complex parser in the netlink sensor** (3-level recursive TLV), and it's implemented correctly with proper validation at each level. Ready to merge.

Reviewed-by: Claude Sonnet 4.5 via agent affa379

PR #183, opened 2026-09-14
sensor-linux-netlink: wire sock_diag listening sockets into schema::Event (issue #92)

old-dov:
## Summary

First slice of #92's `schema`/`correlator`/`tamper` wiring follow-up (the one remaining gap after #179/#182): sock_diag's listening sockets now reach `schema::Event`.

- `schema`: new `Event::ListenPort(ListenPortEvent)` variant. `SCHEMA_VERSION` 10 -> 11, `tests/fixtures/v11/` copied forward from `v10` unchanged + one new `listen_port.json`, per the crate's own golden-fixture discipline.
- `sensor-linux-netlink`: new `listen_port_events()` maps a `snapshot()` to `Event::ListenPort` for every listening socket with an attributable PID.
  - `proc_meta` resolves `comm`/`ppid`/`gid` from `/proc/<pid>/status` (not `/proc/<pid>/stat` — `stat`'s parenthesized `comm` field can itself contain spaces/parens via `prctl` renaming, `status` gives the same values as unambiguous separate lines; confirmed against this dev machine's real `/proc`).
  - `uid` comes straight from the kernel via `sock_diag` itself (`SocketSnapshotEntry::uid`) — no `/proc` read needed for that one, and it's race-free unlike everything resolved from a second, separate `/proc` query.
  - `normalize` does the pure `entry + proc info -> Event` mapping — same module split as `sensor-linux`'s own eBPF-to-schema normalization.
  - A joined PID that fails to resolve (exited between the two `/proc` reads, or another user's process) is silently skipped, not emitted with fabricated metadata.

## Deliberately NOT in this PR (see the crate doc's updated "Deliberately not here yet")

- Conntrack's and proc connector's own schema/`tamper` wiring — conntrack flows carry no PID from the kernel at all (would need a further join, e.g. against a `sock_diag` snapshot's own tuples, to attribute one); proc connector's role per the issue is a `tamper` cross-check (divergence from the eBPF stream) that may not want to be a `schema::Event` at all rather than an internal `crates/tamper` comparison. Both are design decisions beyond what this Linux-only slice should make alone.
- Anything that calls `listen_port_events()` on a cadence and diffs consecutive snapshots for actual drift detection, and anything that feeds its output to `crates/correlator`'s `EventBus` — both need a caller that doesn't exist yet for this crate's data. No lab VM here to validate #92's "listening-port drift" done-when item live, same gap as before this wiring existed.

## Verified

- `cargo test` / `cargo clippy --all-targets -- -D warnings` / `cargo fmt --check` clean for `sensor-linux-netlink` and `schema`.
- `cargo check` clean across every other `schema` consumer in the workspace: `correlator`, `enrich`, `rules`, `sigma`, `sinks`, `store`, `yara`, `sensor-linux-lsm`, `sensor-linux`. `ml` excluded — its `ort`/`openssl-sys` build dependency needs `pkg-config`, which this dev machine doesn't have; confirmed via `cargo tree -i openssl-sys` that the dependency path is unrelated to this change.


SherlockOmss (approved):
## Review: APPROVED

Comprehensive review completed (agent a8e44ab). Correct implementation with no issues found.

### Key Strengths

**Schema Version Bump (v10→v11):**
- SCHEMA_VERSION correctly bumped 10→11
- New ListenPortEvent struct properly added
- Event::ListenPort variant added to enum
- All v10 golden fixtures copied to v11/ unchanged (verified via diff)
- New v11/listen_port.json fixture added
- Event::meta() updated to handle new variant
- All tests updated and passing

**/proc/<pid>/status vs /proc/<pid>/stat - CORRECT CHOICE:**
- Uses /proc/<pid>/status with separate Name:/PPid:/Gid: lines
- Avoids /proc/<pid>/stat's ambiguous parenthesized comm field
- Correctly identified prctl(PR_SET_NAME) renaming issue
- More robust parsing with unambiguous field boundaries

**PID Resolution Failure Handling:**
- Failed resolution silently skipped (no fabricated metadata)
- Correctly handles: process exit race, EACCES on other users, malformed proc
- filter_map pattern ensures only successful resolutions reach events
- Documented as "best-effort per-poll sample"

**Event Normalization:**
- UID from kernel sock_diag (authoritative, race-free)
- GID from /proc (best-effort, comment notes setgid race window)
- Container context None (sock_diag has no cgroup access, correctly deferred)
- Only Listen state sockets emit events (Established filtered out correctly)

**Breaking Changes:**
- NONE - Event enum is #[non_exhaustive]
- All schema consumers compile clean (correlator, rules, sigma, sinks, store)
- Event::meta() handles all variants

### Verification

- Dependency rules: PASS (38 crates)
- All schema consumers: PASS (cargo check)
- Golden fixtures: COMPLETE (all v10 copied + new listen_port.json)
- Tests: Unit + integration coverage

### Scope Discipline

Correctly deferred (documented in crate doc):
- Conntrack/proc connector schema wiring (needs design decisions)
- Polling cadence/drift detection (caller responsibility)
- correlator::EventBus integration (no caller exists)
- UDP sockets (different semantics)

All tracked as #92 follow-ups, not silently dropped.

**Recommendation:** Ready to merge. Clean implementation with appropriate error handling and comprehensive test coverage.

Reviewed-by: Claude Sonnet 4.5 via agent a8e44ab

PR #184, opened 2026-09-14
Fix: sensor-windows Capabilities missing auth_events field

SoFrexs:


old-dov:
Reviewed. **This does not compile as-is:**

```
crates\sensors\windows\etw\src\sensor.rs:225:13: error[E0560]: struct `Capabilities` has no field named `auth_events`: `Capabilities` does not have this field
```

Confirmed by building natively on Windows (`cargo check -p sensor-windows`). Note this module is `#[cfg(windows)]`-gated, so a `cargo check` on WSL/Linux silently skips it entirely and won't catch this — that's presumably how it slipped through local pre-push checks.

The PR only touches `crates/sensors/windows/etw/src/sensor.rs`, adding `auth_events: false` to the `Capabilities` struct literal — but `schema::Capabilities` (`crates/schema/src/sensor.rs`) itself was never given that field. It needs a companion change there:

```rust
pub struct Capabilities {
    ...
    pub auth_events: bool,
    ...
}
```

Once that field exists, two other exhaustive struct literals will need it too or they'll break with E0063 (missing field) instead of building:
- `crates/sensors/linux/userspace/src/sensor.rs:474` (`LinuxSensor::capabilities`)
- any conformance-suite fixtures that construct a `Capabilities` by hand

Both current implementors list every field explicitly (no `..Default::default()`), so this is a hard compile break across the workspace, not just this crate — worth fixing all three call sites together in this PR.

Minor nit once it compiles: the doc comment says "see that crate's `EventLogSensor::capabilities` for the true value" — `sensor-windows-eventlog` is currently just an 11-line stub, no `EventLogSensor` type exists yet. Fine as a forward reference to planned work, just flagging so nobody greps for it and gets confused in the meantime.

SoFrexs:
Subsumed by #188 (merged), which adds auth_events to schema::Capabilities and fixes all call sites. Closing per Jean's review.

Issue #112, opened 2026-09-03
Watchdog service hardening: Windows args persistence, systemd quoting, relative paths


florianamette:
Review findings on the watchdog PRs:
- Windows service registration does not persist the agent's arguments — after reboot the service restarts the agent with defaults, silently dropping the configured alerts/events paths.
- systemd unit generation does not quote paths (a space in the path splits the ExecStart).
- Alert/restart-log paths are cwd-relative; under a service manager cwd is / (same class as the agent content-path fix in PR #106 — resolve against the executable or make them absolute at install time).

SoFrexs:
Fixed: agent-bin/alerts persisted through the service's binPath instead of relying on service_main's args.

Issue #102, opened 2026-09-03
Watchdog: liveness/heartbeat health check (detect a hung agent)


florianamette:
**Component:** `watchdog/` · supervision loop; likely needs `crates/ipc`.

The watchdog only checks whether the agent **process** is alive (`try_wait`). A hung agent — deadlocked, wedged on a poisoned lock, event loop stalled, sensor thread dead but main thread parked — counts as perfectly healthy and is never restarted. For an EDR that is the worst failure mode: the endpoint looks protected while collecting nothing.

Add a real liveness check so "the agent is running" means "the agent is doing its job":

- The agent emits a heartbeat the watchdog can observe (IPC ping/pong, a touched liveness file with a monotonically advancing counter/timestamp, or a socket the watchdog polls). Reuse `crates/ipc` rather than inventing a channel.
- The watchdog treats a missed heartbeat (N intervals) as a failure: kill + restart, same path as a crash, feeding the backoff (#101).
- The heartbeat should reflect real progress (sensor pipeline advancing), not just "main thread scheduled a timer", so a wedged sensor is caught.
- Bound it so a legitimately busy agent isn't killed under load — tune interval/threshold conservatively and make them configurable.

Cross-platform: the mechanism must work identically on Windows/Linux/macOS (no OS-specific liveness primitive in the core loop).

## Done when
- [ ] Agent emits a progress-backed heartbeat; watchdog consumes it
- [ ] A deliberately hung agent (test hook that stops pinging) is detected and restarted
- [ ] A busy-but-healthy agent is not falsely killed (load test / generous threshold)
- [ ] Interval + miss-threshold are configurable; defaults documented

SoFrexs:
Fixed: progress-backed heartbeat file the watchdog polls; kills/restarts a stalled agent into the #101 backoff.

Issue #101, opened 2026-09-03
Watchdog: exponential backoff on agent crash loops


florianamette:
**Component:** `watchdog/` · supervision loop in `watchdog_loop`.

The supervision loop restarts the agent on a fixed `restart_delay` (default 5s) forever. An agent that crashes on startup — bad config, missing rules content, a panic on a corrupt store — becomes a tight 5s respawn loop: CPU/log/journal churn, and on Windows/systemd/launchd the outer service-manager layer piles its own restarts on top. There is no ceiling and no signal that the agent is failing rather than being killed.

Add crash-loop dampening so persistent failure degrades gracefully instead of hammering:

- Exponential backoff with a cap (e.g. 5s → 10s → 20s … up to a few minutes) when the agent exits quickly after start.
- Reset the backoff once the agent has stayed up past a "healthy" threshold (so a one-off crash after a week doesn't inherit a long delay).
- Distinguish a fast crash (exited < N seconds after spawn) from a normal long-lived exit — only the former should back off.
- Keep the existing `restart_delay` as the floor; the backoff multiplies from it.

Coordinate with the service-manager restart policies (`sc failure`, `RestartSec`, launchd `ThrottleInterval`) so the two layers don't fight — the watchdog owns agent-level backoff, the service manager owns watchdog-level restart.

## Done when
- [ ] A crash-looping agent backs off exponentially up to a cap instead of restarting every 5s
- [ ] Backoff resets after the agent stays healthy for the threshold
- [ ] Unit test drives the backoff schedule (fast-crash vs. clean long exit) with a fake clock
- [ ] Behaviour documented in the module doc alongside the two-layer model

SoFrexs:
Fixed: exponential backoff on fast crash loops, resetting once the agent runs past a healthy threshold.

PR #185, opened 2026-09-14
Watchdog: crash-loop backoff (#101), arg persistence fix (#112), heartbeat liveness (#102)

SoFrexs:
- **Watchdog: fix Windows service arg persistence (#112), add crash-loop backoff (#101)**
- **Watchdog: heartbeat-based liveness detection for hung agents (#102)**


old-dov:
Reviewed #185 in full — solid design. `Backoff` and `HeartbeatMonitor` as pure state machines (fed hand-picked values instead of real sleeps/real files) makes both fully unit-testable, and the #112 Windows `binPath` persistence fix is a genuine and subtle bug caught correctly: `service_main`'s argument list really is only populated when a caller explicitly invokes `StartService` with args, never on ordinary SCM auto-start at boot — reading `std::env::args()` back instead is the right fix, since the SCM does launch the process with the literal `binPath=` command line every time, auto-start included.

Compiled and ran the full suite natively on Windows (not WSL, so the `cfg(windows)` service code is actually exercised): `agent` 8/8 pass, `watchdog` 18/19 pass. One real, confirmed failure:

```
thread 'supervise::tests::tick_only_checks_the_file_once_per_interval' panicked at watchdog\src\supervise.rs:468:69:
called `Result::unwrap()` on an `Err` value: Os { code: 3, kind: NotFound, message: "Le chemin d'accès spécifié est introuvable." }
```

`watchdog/src/supervise.rs`'s new test hardcodes `/tmp/heartbeat-monitor-tick-test.txt`, which doesn't exist on Windows. `agent/src/heartbeat.rs`'s tests in this same PR already use `std::env::temp_dir()` for exactly this reason — `supervise.rs`'s test should do the same:

```rust
let path = std::env::temp_dir().join(format!("heartbeat-monitor-tick-test-{}.txt", std::process::id()));
```

Given #112 in this very PR is specifically about a Windows-only footgun, this one is worth fixing before merge rather than after — `cargo test -p watchdog` currently fails outright on Windows.

Clippy clean on both crates (`-D warnings`). No other findings — just that one test needs the temp-dir fix.

old-dov (approved):
Rebased onto main cleanly, mergeable. LGTM.

Sollykhan (approved):
Re-review post-rebase. Vérifié :
- Rebase propre : 4 commits linéaires ordonnés (#202 → #101+#112 → #102 → fix test Windows)
- Le 'removal' apparent de `account_creations_enabled` sur windows.rs vs origin/main est un artefact d'affichage : mergeStateStatus CLEAN confirme que le 3-way merge au squash préservera les deux additions (regions différentes du fichier)
- Ajustement de `explicit_agent_path_wins` cohérent avec `explicit_relative_agent_path_is_made_absolute` — bon pattern pour absorber la fix #202 sans casser la sémantique du test
- `heartbeat_path_for` bien documenté avec sa contrainte de parité vs `agent::heartbeat::heartbeat_path_for`, et un test qui l'asserte
- `cargo test -p watchdog` : 15/15 en local

LGTM, prêt à merger.

Issue #123, opened 2026-09-04
Alpine lab target: musl/BusyBox validation


florianamette:
Add an Alpine Linux target to the lab harness (`lab/vagrant`, `lab/provisioning`, `lab/MATRIX.md`), alongside the existing Ubuntu box.

**Why Alpine specifically** — it exercises paths the glibc/Ubuntu box cannot:
- **musl libc + BusyBox userland**: the agent and its `libc` FFI (`geteuid`, `setrlimit`, `clock_gettime` in the eBPF loader; `/proc` parsing) must work against musl, not just glibc. This is the closest cheap proxy for "does the static agent run on a minimal host."
- **Different `/proc` and coreutils shapes**: `seed_from_proc`, the capability check (`/proc/self/status` CapEff), and the download-then-exec scenarios assume GNU coreutils behavior; BusyBox `wget`/`sh`/`nc` differ (see the #113 provisioning fixes — Alpine widens that surface).
- **eBPF on Alpine's kernel**: validates the bpf-linker LLVM-major alignment and BTF presence on a distro that ships its own kernel flavor.

**Scope**
- A `lab/vagrant` Alpine box (or container target if a VM is impractical on Apple Silicon — note the choice in `lab/MATRIX.md`).
- Provisioning that installs the eBPF toolchain + scenario tools via `apk` (BusyBox `nc`/`wget` caveats from #113 apply).
- The walking-skeleton scenario (beacon) must fire end-to-end, same acceptance as the Ubuntu box.

Relates to #10 (lab harness), #34 (audit fallback — Alpine is a good second-sensor host), #113 (provisioning portability), #53 (CO-RE across kernels).

Issue #113, opened 2026-09-03
Lab provisioning fixes: nc install, Debian linux-perf, listener cleanup, bpf-linker revalidation


florianamette:
Review findings on the lab PRs:
- scenarios use `nc` but provisioning never installs it (fresh minimal images fail the beacon scenario);
- Debian ships `linux-perf`, not `linux-tools-*` (Ubuntu name) — provisioning fails on Debian boxes;
- the scenario teardown kills by pattern and can miss the actual nc child (or match unrelated processes — the pkill -f incident from VM validation);
- bpf-linker is not revalidated when the pinned nightly (and its LLVM major) bumps — provisioning must re-check the LLVM-major alignment (lab/provisioning/linux-toolchain.sh).

PR #186, opened 2026-09-14
lab: Alpine provisioning script and matrix entry (#123)

Jihair54:
## Summary
- Adds `lab/provisioning/alpine-toolchain.sh`, the apk-based companion to `linux-toolchain.sh`, verified end-to-end on a real Alpine 3.24 VM (system packages, rustup toolchain, bpf-linker, full agent build/test/clippy with the eBPF sensor loaded and all 5 programs accepted by the verifier).
- Adds the `alpine-3.24` row to `lab/MATRIX.md`.

## Key finding
`cargo install bpf-linker` on musl produces a broken static/dynamic hybrid binary (no `PT_INTERP` segment) that SIGSEGVs on its first LLVM call. The upstream prebuilt musl-static release — the one `linux-toolchain.sh` already downloads for the other distros — sidesteps this entirely and needs no LLVM-major alignment with the pinned nightly. Verified against both the pinned nightly (LLVM 22) and a plain `rustup toolchain install nightly` (LLVM 23).

## Honest limitations (not done here)
- No `lab/vagrant/Vagrantfile` entry. The existing Vagrantfile targets QEMU/arm64 on an Apple Silicon host; this work was validated on VirtualBox/x86_64 from Windows, so a Vagrant box choice for Alpine on that provider/arch is unverified — adding an untested entry seemed worse than leaving it out.
- `lab/scenarios/beacon.sh` (the official walking-skeleton scenario) was not replayed as-is; capture was validated manually (`agent capture-events` against real activity) instead. BusyBox `wget`/`sh`/`nc` differences (same class of issue as #113) are noted as a caveat in the script's final output but not resolved.

## Test plan
- [x] `cargo build --release --workspace --exclude sensor-linux-ebpf` on Alpine 3.24
- [x] `cargo test --workspace --exclude sensor-linux-ebpf --release` — 79 test groups, 0 failed
- [x] `cargo clippy --workspace --exclude sensor-linux-ebpf --all-targets -- -D warnings` — clean
- [x] `python3 tools/check-deps.py` — clean
- [x] `agent status` — all 5 eBPF programs accepted by the verifier
- [x] Re-ran `alpine-toolchain.sh` itself on the already-provisioned VM — idempotent, no errors
- [ ] `lab/scenarios/beacon.sh` replayed as-is (not done, see above)
- [ ] `vagrant up` on an actual Vagrant/Alpine box (not done, see above)

Closes partially #123 — the toolchain/provisioning half is done and verified; the Vagrant harness integration and official scenario replay are left as follow-up.


Jihair54:
Update: replayed the official `lab/scenarios/beacon.sh` walking-skeleton scenario as-is on the Alpine VM (not just manual capture as noted in the original description).

`sudo target/release/agent run` + the scenario script produced exactly the documented acceptance alert:

```
[ALERT] T1071/T1041 — pid=29441 comm=nc → 127.0.0.1:4444 | 3x in 60s — suspected beaconing
```

Also saw the correlator's respawn detection (`T1059/T1071`) and a Bayesian alert on background `chronyd` NTP activity fire correctly alongside it — the full detection pipeline (sensor, rules, correlator) is working end-to-end on Alpine, not just the sensor in isolation.

One correction to the script's caveat about BusyBox `nc`: this VM has real OpenBSD netcat installed (`nc -h` reports "OpenBSD netcat (Debian patchlevel...)"), not BusyBox's built-in. The BusyBox-specific `wget`/`sh`/`nc` differences noted in the script are therefore still unverified on a genuinely minimal fresh Alpine image — leaving that caveat in place as-is.

Still not done: no `lab/vagrant/Vagrantfile` entry (unverified provider/arch mismatch, as noted above).

old-dov:
Reviewed `lab/provisioning/alpine-toolchain.sh` + the `MATRIX.md` entry — good work, matches `linux-toolchain.sh`'s conventions (idempotent, `set -euo pipefail`, same stale-build-cache-bust helper for the `sensor-linux-*` fingerprint dirs). The musl `bpf-linker` workaround (prebuilt static release instead of `cargo install`, which hits a real musl PT_INTERP linking bug) is well-justified, and the caveats section at the end is honest about what's NOT verified (bindgen/aya-tool, `lab/scenarios` BusyBox compat, `ml`'s onnxruntime gap).

One loose end: the top doc comment points to `BpfLinker_Alpine_Crash_Explique.md at the repo root` for "the full source-build post-mortem" — that file doesn't exist anywhere in this PR or in `main`'s history (checked `git log --all -- '*BpfLinker*'`, nothing). Either it needs to land alongside this PR, or the comment should drop the dangling pointer (or say where the writeup actually lives — Discord?). Also, if it does get added: the filename mixes French into an English-only repo (commits/docs/PRs are English-only per team convention) — worth an English name from the start.

No blocking issues from a read-through; didn't have a spare Alpine VM handy to run it end-to-end myself, so I'm trusting the "verified end-to-end on a real Alpine 3.24 VM" note in the commit message.

Jihair54:
Good catch — that file was a local, untracked note from my own learning notes, never meant to be part of the repo (and yeah, in French, so it wouldn't have belonged as-is anyway). Dropped the dangling pointer in 3d0aa54; the comment now points at this PR discussion instead, which is where the actual three-bug post-mortem lives if it's ever needed again.

Jihair54:
Closed one of the two unverified caveats from the script's original header: `bindgen-cli` + `aya-tool` regeneration, now actually exercised on the Alpine VM (not just skimmed).

`bindgen-cli` hit exactly the class of bug this whole script is built around — `cargo install bindgen-cli` produces a static-pie binary by default, and static musl binaries can't `dlopen()` at all, so it panicked reaching for `libclang.so` ("Dynamic loading not supported") regardless of path. Same fix as bpf-linker/ort: rebuild with `RUSTFLAGS="-C target-feature=-crt-static"`. Script updated (47025b7).

`aya-tool` needed nothing — it reads BTF directly rather than going through libclang/dlopen, so it works fine static-pie out of the box. Ran `aya-tool generate task_struct` against this VM's kernel BTF and got valid bindgen output.

Remaining honest gap from the original caveats list: `lab/scenarios/*.sh`'s BusyBox-vs-coreutils assumption is still unverified on a genuinely minimal Alpine image (this VM has real coreutils/netcat installed, not BusyBox's).

Jihair54:
Second and last caveat closed. Turns out this VM had `netcat-openbsd` explicitly installed (not part of this script — must have been added manually earlier for testing), which shadowed BusyBox's own `nc` applet at the same path. Uninstalled it, confirmed `/usr/bin/nc -> /bin/busybox` (real BusyBox now), and reran `lab/scenarios/beacon.sh` end to end against a live `agent run`.

Works correctly: the listener's `nc -l -p PORT || nc -l PORT` fallback (already in the script, from #113) picks the right syntax for BusyBox, and the `T1071/T1041 — ... suspected beaconing` alert fires exactly as documented, same as the earlier run with real netcat.

Pushed in f41ed02. Both caveats from the script's original list are now verified rather than assumed — the only thing genuinely left on #123 is the Vagrantfile integration, which stays out of reach from this environment (provider/arch mismatch, noted from the start).

PR #193, opened 2026-09-15
sensor-linux-netlink: wire conntrack beacon + listener-drift detection end-to-end (issue #92)

old-dov:
## Summary

Closes 2 of issue #92's 3 done-whens (the 3rd, the proc-connector/`tamper` cross-check, is explicitly out of scope — see below):

**1. "Conntrack features reach the correlator for a beacon scenario"**
- **`schema::NetworkFlowEvent` gains `local_port`** (`SCHEMA_VERSION` 12→13, `v13/` fixtures). Already computed internally in `sensor-linux-netlink` (the sock_diag join) but wasn't exposed. Needed because a poll-based source re-reports the *same* open flow on every poll, unlike a discrete `connect()` trace — without a stable per-flow identity, a naive counter would false-positive BEACON on any ordinary long-lived connection (SSH, a websocket) simply for still being open on its 3rd poll inside the 60s window.
- **`crates/rules::check_beacon_flow`** (new, alongside the existing `check_beacon`): same T1071/T1041 rule, fed by `NetworkFlowEvent` instead of `ConnectEvent`, deduped by `local_port` (`FlowPortDedup`, new in `sliding.rs`). Shares the same underlying counter as `check_beacon` — a mixed source (eBPF `Connect` + netlink `NetworkFlow` both observing the same real connection) still crosses the threshold correctly rather than resetting it (covered by a dedicated test).
- **`crates/correlator::is_correlated`** now includes `NetworkFlow`, pushing it to the bus. Deliberately **no new co-occurrence rule or `BehaviorVector` feature** yet — documented in `event.rs`: folding it into the `Connect`-keyed rules risks double-alerting when both sources see the same connection, and touching `BehaviorVector` risks shifting ML-calibrated feature values without Nikolas.

**2. "Listening-port drift on a lab VM produces events"**
- **`crates/rules::check_listen_port_drift`** (T1571 — closest existing tag in this crate, no better precedent here for "a new listener appeared"): fires once per `(local_addr, local_port)` not in the startup baseline and not already alerted this run.
- **`RuleState::seed_listen_ports`**, called from `seeded_rule_state()` via one startup `sock_diag` snapshot — every listener already up when the agent attaches (sshd, nginx from systemd) is the baseline, not a finding, same principle as `seed_from_proc`.
- No name/path exclusion list yet, unlike BEACON's `BROWSERS`/`STANDARD_PORTS` — documented as a known gap: no lab capture here to calibrate one against, so a dev server or `docker-proxy` binding a fresh port after startup will alert. Left for real lab data, not guessed at.

**The caller that made both possible:** `sensor-linux-netlink` was **never actually invoked by `agent run`** — no polling thread existed anywhere (its own crate doc says so: "a caller (not yet written)"). `agent/src/commands/linux.rs` now spawns one (`sock_diag` + conntrack, 10s interval), sharing the `DetectionSink` with the eBPF sensor via a new `impl<T: EventSink> EventSink for Arc<T>` in `schema::sensor`. Logs are deduped per-source so an unprivileged agent doesn't spam identical `EPERM` warnings every 10s forever.

## Out of scope

Issue #92's 3rd done-when (proc-connector cross-check against the eBPF stream, divergence = `tamper` signal) is left untouched — its own crate doc already flags this as a cross-team design call ("may not want to be a `schema::Event` in the first place"), not something to decide solo.

## Testing

- 8 new tests in `crates/rules` (`tests::linux::beacon_flow_*` + `listen_port_*`): beacon true-positive/false-positive-avoidance/standard-port/mixed-source, and listener unseeded-alerts/seeded-baseline/repoll-no-realert/two-distinct-alert.
- Full `schema`/`sensor-linux-netlink`/`rules`/`correlator`/`agent` suites green, `cargo clippy -D warnings` and `cargo +nightly fmt --check` clean.
- Attempted a live-fire validation against `lab/scenarios/beacon.sh` with a disposable probe binary (not part of this diff) — inconclusive on this WSL2 box: conntrack accounting stays empty without an nftables/iptables rule referencing conntrack, which this lab has none of (`iptables -L`/`nft list ruleset` both confirmed empty) — a gap already documented from a prior session, not something touched here. `sock_diag`'s live tests (listen-port drift's own data source) are unaffected and pass as before.
- The pre-existing `conntrack_flow_events_runs_end_to_end_against_the_real_kernel_and_proc` failure (`NLMSG_ERROR -1`) reproduces identically on unmodified `main` in this environment — confirmed not a regression before pushing either commit.


Sollykhan (approved):
Reviewed — approving.

On the deliberate ML-side hold: **fully agree**. Adding `NetworkFlow` to `BehaviorVector` now would shift the feature space that the current `cmdline-iforest-*/0.1.0` and any downstream T1/T2 models were calibrated against, silently changing scores on every existing baseline without a retrain to match. The right sequence is (1) land this PR with the new events flowing, (2) capture fresh benign baselines that include netlink-sourced flows (my linux-wsl2-solkapc-2026-09-09-v2 baseline predates this wiring, so it won't help), (3) decide the feature contract for `NetworkFlow` in `BehaviorVector` explicitly, (4) retrain. Same reasoning for the `Connect`-keyed correlation rules — double-alerting on mixed sources is a real risk. Good instinct to hold.

On the scope: 2/3 done-whens with the 3rd (proc-connector tamper cross-check) explicitly deferred as a cross-team design call is the right call. Worth opening a follow-up issue for it so the deferral is tracked in GitHub, not just in the PR body.

On the listener drift name/path exclusion list gap: same logic — waiting for real lab data rather than guessing is correct. Leaves the door open for a follow-up once the lab captures pick up representative noise.

Rest of the PR (schema v13 bump with `local_port`, `FlowPortDedup`, sensor wiring via `EventSink for Arc<T>`, per-source log dedup): reads clean, tests cover the tricky cases (mixed-source counter sharing especially).

LGTM.

Sollykhan (approved):
Reviewed — approving, leaving the merge call to you.

On the deliberate ML-side hold: **fully agree**. Adding `NetworkFlow` to `BehaviorVector` now would shift the feature space that the current `cmdline-iforest-*/0.1.0` and any downstream T1/T2 models were calibrated against, silently changing scores on every existing baseline without a retrain to match. The right sequence is (1) land this PR with the new events flowing, (2) capture fresh benign baselines that include netlink-sourced flows (my linux-wsl2-solkapc-2026-09-09-v2 baseline predates this wiring, so it won't help), (3) decide the feature contract for `NetworkFlow` in `BehaviorVector` explicitly, (4) retrain. Same reasoning for the `Connect`-keyed correlation rules — double-alerting on mixed sources is a real risk. Good instinct to hold.

On the scope: 2/3 done-whens with the 3rd (proc-connector tamper cross-check) explicitly deferred as a cross-team design call is the right call. Worth opening a follow-up issue for it so the deferral is tracked in GitHub, not just in the PR body.

On the listener drift name/path exclusion list gap: same logic — waiting for real lab data rather than guessing is correct. Leaves the door open for a follow-up once the lab captures pick up representative noise.

Rest of the PR (schema v13 bump with \`local_port\`, \`FlowPortDedup\`, sensor wiring via \`EventSink for Arc<T>\`, per-source log dedup): reads clean, tests cover the tricky cases (mixed-source counter sharing especially).

LGTM.

PR #187, opened 2026-09-14
ml/data: CLI wrapper for manifest read/write

Sollykhan:
## Summary

Adds `python -m synthaea_ml.data.manifest {write,verify}` — the CLI wrapper flagged as a "known limitation" in #181. Follow-up to Jean's review: the manual friction of the Python one-liner had a 0/2 completion rate on the current baselines, so wrapping it stops being optional.

## What it does

- `write` takes `--platform`, `--os-version`, `--workload-label`, `--capture-start`/`-end` (ISO-8601 with timezone), plus optional `--capturer-ref`, `--baseline-filename`, `--hostname`.
- `verify` wraps `verify_manifest`.
- `--platform` and `--workload-label` enforce `choices=` so a typo does not land silently in a shipped manifest.
- `--baseline-filename` is exposed so a legacy capture keeping its original name (`baseline_benign.jsonl`) can get a manifest without renaming first.

## Tests

13 new tests in `test_manifest_cli.py`:
- 3 write happy paths (default filename, `--baseline-filename` override, offset TZ normalisation)
- 4 write error paths (unknown platform, unknown workload_label, naive datetime, missing baseline)
- 3 verify (happy, mutation detected, missing manifest)
- 3 unit tests for `_parse_capture_time`

Existing 17 tests in `test_manifest.py` still pass.

## Not in this PR

- **`verify_manifest` bug**: it hardcodes `DEFAULT_BASELINE_FILENAME` in its file lookup, so `python -m synthaea_ml.data.manifest verify` on a legacy `baseline_benign.jsonl` fails with `FileNotFoundError` even after `write` succeeded. Fix belongs to a follow-up — either store the filename in the manifest (schema bump v2) or take it as a `verify_manifest` parameter. Preference: parameter, no schema churn.
- **Migration of the two existing baselines** (`linux-wsl2-solkapc-2026-09-09`, `windows-solkapc-2026-09-08`) — separate data-migration PR once the `verify_manifest` fix lands.

## Test plan

- [x] `ruff check` clean
- [x] `pytest ml/tests/test_manifest_cli.py -v` → 13 passed
- [x] `pytest ml/tests/test_manifest.py -v` → 17 passed (unchanged)
- [x] Manual `python -m synthaea_ml.data.manifest --help` → parser prints subcommands correctly

## Related

Part of #44. Follow-up to Jean's review on #181 (comment on GitHub). Unblocks #181 for merge once this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #194, opened 2026-09-15
ml/data: let verify_manifest check a non-default baseline filename

### LINKED ISSUE #190: ml/data: verify_manifest hardcodes DEFAULT_BASELINE_FILENAME
## Problem

`verify_manifest(baseline_dir)` in `ml/synthaea_ml/data/manifest.py` looks up the sample file under `DEFAULT_BASELINE_FILENAME` (`"baseline.jsonl"`) in hardcoded form:

```python
def verify_manifest(baseline_dir: Path) -> None:
    manifest = load_manifest(baseline_dir)
    baseline_path = baseline_dir / DEFAULT_BASELINE_FILENAME    # <-- here
    if not baseline_path.exists():
        raise FileNotFoundError(f"baseline not found: {baseline_path}")
    ...
```

But `write_manifest` accepts a `baseline_filename` parameter (default `"baseline.jsonl"`, overridable e.g. to `"baseline_benign.jsonl"` for legacy captures). Asymmetry: you can write a manifest for a legacy file name, but you can't verify it.

## Impact

- `python -m synthaea_ml.data.manifest verify <legacy-dir>` fails with `FileNotFoundError` even after `write` succeeded.
- Blocks the migration of the two existing baselines (`linux-wsl2-solkapc-2026-09-09`, `windows-solkapc-2026-09-08`) — they keep the sensor's original `baseline_benign.jsonl` name.

## Options

1. **Store `baseline_filename` in the manifest** (schema_version bump to 2). Cleaner long-term but every reader has to handle both v1 and v2.
2. **Take `baseline_filename` as a `verify_manifest` parameter**, defaulting to `DEFAULT_BASELINE_FILENAME`. No schema churn. CLI would take a `--baseline-filename` on `verify` too.

Preference: **option 2**. Small change, no cross-version compat cost, unblocks the two legacy migrations immediately.

## Related

- Introduced by #170 (baseline manifest module).
- Surfaced by #187 (CLI wrapper — the CLI can `write` but not `verify` legacy names).
- Blocks the docs correction in #181.

Jihair54:
## Summary
- `write_manifest` accepts a `baseline_filename` override for legacy captures that kept their original file name (e.g. `baseline_benign.jsonl`), but `verify_manifest` hardcoded `DEFAULT_BASELINE_FILENAME` in its lookup — a manifest written that way could never be verified (`FileNotFoundError`).
- Adds the same keyword-only `baseline_filename` parameter to `verify_manifest`, defaulting to `DEFAULT_BASELINE_FILENAME` (option 2 from the issue — no schema churn).

## Not in this PR
- The CLI (`python -m synthaea_ml.data.manifest verify --baseline-filename ...`) isn't wired up here — that subcommand parser is introduced by #187, which hasn't merged yet. This PR only fixes the library function; the CLI flag is a trivial follow-up once #187 lands.
- Migration of the two existing legacy baselines (`linux-wsl2-solkapc-2026-09-09`, `windows-solkapc-2026-09-08`) — separate PR per #190's description.

## Test plan
- [x] `pytest ml/tests/test_manifest.py -v` → 19 passed (2 new: accepts a legacy filename when both write/verify agree on it, and reports the *correct* missing file when verify still uses the default against a legacy baseline)
- [x] `pytest ml/tests -q` → 82 passed (full `ml/` suite, no regressions)
- [x] `ruff check` clean
- [x] Checked all existing callers (`registry/training_record.py`, both call sites) — single positional arg, unaffected by the new keyword-only parameter

Fixes #190.


Sollykhan (approved):
Nice work @Jihair54. Option 2 from #190 exactly as preferred — parameter over schema bump.

Both new tests hit the right notes: happy path against a legacy filename, and the wrong-filename case that fails with an actionable message. The second one is a nice signal — you're not just checking that a bad call fails, you're checking that it fails informatively.

Discipline elsewhere too: callers checked (`registry/training_record.py`), +46/-2 surgical, scope kept clean (CLI wiring deferred).

Now that #187 is merged, wiring up `--baseline-filename` on the CLI's `verify` subcommand is ~10 lines against the existing argparse setup. Take it if you want to extend — good excuse to get familiar with subparsers.

LGTM.

PR #195, opened 2026-09-15
ml/export: dispatch verify_onnx sanity samples by platform

### LINKED ISSUE #191: ml/export: verify_onnx uses Windows sanity samples for every model
## Problem

`ml/synthaea_ml/export/verify_onnx.py` imports `SANITY_CHECK_SAMPLES` from `train_windows` and feeds it to every model, regardless of platform:

    from synthaea_ml.training.train_windows import SANITY_CHECK_SAMPLES

The Windows samples are Windows cmdlines (`svchost.exe`, `powershell -EncodedCommand …`, `AppData/…`). A Linux model scored against them produces uniformly negative scores — everything looks anomalous because none of these tokens belong in the Linux feature space.

## Impact

Cosmetic — the runtime property checks (finiteness, determinism, label/score consistency) and the provenance check that gate ship are correct. But the logs are misleading and hide the actual sanity signal on Linux models:

    registry/cmdline-iforest-linux/0.1.0/model.onnx: input X, outputs ['label', 'scores']
      benign (svchost): score=-0.0166 -> ANOMALY
      benign (powershell): score=-0.0972 -> ANOMALY
      suspicious (AppData): score=-0.0783 -> ANOMALY
      suspicious (base64 ps): score=-0.0513 -> ANOMALY
      suspicious (Temp): score=-0.0222 -> ANOMALY

## Options

1. **Dispatch on the model's training.json** — look at `dataset_versions[*].platform` and load the matching platform's `SANITY_CHECK_SAMPLES`. Robust but assumes provenance is present (fine going forward, not for legacy).
2. **Dispatch on the model dir path** — `cmdline-iforest-linux/` → linux samples, `cmdline-iforest-windows/` → windows samples. Brittle but works for legacy too.
3. **Fall back to (2) when no training.json**, use (1) otherwise.

Preference: **option 3**.

## Related

- Surfaced by #180 (release gate hook) manual test on `cmdline-iforest-linux/0.1.0/model.onnx`.
- Documented as a known limitation in `docs/ml/data-infra.md` (#181).

Jihair54:
## Summary
- `verify_onnx.verify()` fed every registry model the Windows `SANITY_CHECK_SAMPLES`, regardless of the model's actual platform. A Linux model scored against Windows cmdlines (`svchost.exe`, `powershell -EncodedCommand`, `AppData/...`) produces uniformly negative scores — misleading logs, even though the runtime property checks that actually gate ship (finite, deterministic, label/score agreement) stay correct.
- Adds `sanity_samples_for(model_dir)`, implementing option 3 from the issue: prefer the platform recorded in `training.json`'s `dataset_versions[0].name` (`<platform>__...` convention from `default_dataset_name`), fall back to the legacy registry directory-name convention (`cmdline-iforest-linux/` -> linux) for pre-#44 entries that have no training record. Raises `VerificationError` rather than silently defaulting when neither signal resolves.

## Test plan
- [x] `pytest ml/tests/test_verify_onnx.py -v` → 6 passed (legacy dir dispatch both ways, training-record dispatch, **training record wins over a disagreeing directory name** — the precedence this fix exists for — unknown-platform fallback, and the fail-closed case)
- [x] `pytest ml/tests -q` → full suite green, no regressions
- [x] `ruff check` clean
- [x] Ran `python -m synthaea_ml.export.verify_onnx` against the real registry — before this fix all 5 Linux-model samples showed `ANOMALY`; after, the 3 benign ones (`whoami`, lab script, `modprobe`) correctly show `normal`:
  ```
  registry/cmdline-iforest-linux/0.1.0/model.onnx: input X, outputs ['label', 'scores']
    benign (whoami): score=+0.1625 -> normal
    benign (lab script): score=+0.1008 -> normal
    benign (modprobe): score=+0.0693 -> normal
    suspicious (base64 decode): score=-0.0817 -> ANOMALY
    known FP (curl health): score=-0.0151 -> ANOMALY
  ```

Fixes #191.


Sollykhan (approved):
Nice work again @Jihair54. Option 3 exactly as the issue preferred — dispatch on `training.json` first, dir-name fallback for legacy — and the fail-closed choice (`VerificationError` when neither signal resolves) is the right call over a silent default.

The test that matters most here is `training_record_wins_over_disagreeing_directory_name`. That's the precedence check that will save someone later when a legacy dir gets a manual retrain into a different platform's model space — good instinct to make it explicit rather than trust that the two signals will always agree.

Empirical proof in the PR body (before/after run showing the 3 benign Linux cmdlines flipping from ANOMALY to normal) is the standard everyone should follow — makes the fix undeniable without needing to run it.

LGTM.

Issue #110, opened 2026-09-03
ml crate: ort static linking per ADR-0002


florianamette:
ADR-0002 (revised) decides ort statically linked into the agent. Review findings on #69/#98 note the current ml crate wiring still resolves ort dynamically (load-time dylib dependency), which breaks single-binary deployment and the updater's integrity story. Align the crate features (`ort` with static/compile-from-source config), verify no runtime .so/.dll is required on a clean VM, and record the binary-size delta. Coordinate with the ML workstream — the parity fixtures must pass against the statically linked runtime.

Jihair54:
Proof-of-concept on an (unsupported) Alpine/musl VM: true static linking of onnxruntime into `ml` is achievable, but the current `ort` pin needs manual help to get there.

**Confirmed the gap is not platform-specific.** `ort = "=2.0.0-rc.13"` in the root `Cargo.toml` doesn't opt out of any default features, which include `download-binaries` + `copy-dylibs` — i.e. it downloads a prebuilt *dynamic* onnxruntime and copies it next to the binary at build time. That's the dynamic linking ADR-0002 rejects, and it happens on every platform today, not just Alpine.

**What worked:** built onnxruntime 1.30.0 from source as a static lib (`./build.sh --config Release --update --build --no_telemetry --cmake_extra_defines onnxruntime_BUILD_UNIT_TESTS=OFF ...`, no `--build_shared_lib`), pointed `ORT_LIB_LOCATION` at the `Release` output dir, and ran `cargo test -p ml --release`. All 31 `ml` tests pass, including the ones that run real inference (`scores_match_onnxruntime_reference`, `vectors_match_python_aggregator`). Verified on the resulting test binaries:

```
scorer-80d2c5ddc131c657: ELF 64-bit LSB pie executable, x86-64, ..., static-pie linked, ...
        /lib/ld-musl-x86_64.so.1 (0x...)
```

`file` reports `static-pie linked`, and `ldd` shows no `libonnxruntime.so` line at all (only musl's own PIE loader stub). Zero runtime onnxruntime dependency — the actual goal of the ADR.

**Two build-time gotchas, unrelated to inference itself:**
- `--no_telemetry` is required: without it, onnxruntime builds Microsoft's `cpp_client_telemetry` SDK, which needs `execinfo.h` — a glibc-only header, absent on musl. Disabling telemetry removes the dependency entirely.
[...]

PR #197, opened 2026-09-15
lab/provisioning: script to generate ort static-link RUSTFLAGS (issue #110)

old-dov:
## Summary

Follow-up tooling for issue #110 (ADR-0002: ort statically linked). Not a decision that static linking should ship this way — same "prototype, not a merge" posture as Jihair's original Alpine PoC on that issue.

`lab/provisioning/ort-static-link-flags.sh` walks `$ORT_LIB_LOCATION` for every `.a` file an onnxruntime source build actually produced and emits the `-L`/`-l` RUSTFLAGS needed to statically link it into `crates/ml`, instead of relying on `ort-sys` v2.0.0-rc.13's hardcoded static-link library list — confirmed incomplete for onnxruntime 1.30.0 on both Alpine/musl (Jihair) and Ubuntu 24.04/glibc (my own reproduction, both on issue #110): ~78 abseil sub-libraries (the `Cord`/`Cordz`/`Status` family) plus `utf8_range` and `model_package` genuinely get linked into onnxruntime's own static libs but sit in directories `ort-sys` never searches.

Also documents the one Ubuntu-specific extra step: onnxruntime's CPU provider only pulls in `re2`'s headers (`onnxruntime_add_include_to_target`), not a real `target_link_libraries` — without a shared-lib or test target to create a CMake dependency on it, `re2` is never scheduled to build at all. `cmake --build --target re2` explicitly, before running this script.

## Testing

Verified end-to-end against the Ubuntu build from #110's investigation:
```
export ORT_LIB_LOCATION=.../onnxruntime/build/Linux/Release
eval "$(./lab/provisioning/ort-static-link-flags.sh)"
cargo test -p ml --release
```
Clean 31/31, including real inference (`scores_match_onnxruntime_reference` on both `correlation_scorer.rs` and `scorer.rs`), `ldd` on the resulting test binary confirms zero onnxruntime dylib dependency. `shellcheck` clean.

Nothing else touched — the workspace `ort` dependency, CI, and Cargo.toml are unchanged. This is opt-in tooling for whoever continues #110, not something anything else calls.


Jihair54:
@old-dov Nice — this closes the exact open question I left on #110 ("worth checking if Ubuntu hits the same ort-sys library-list gap"). Good to have it confirmed cross-platform rather than assumed to be an Alpine/musl quirk, and the re2-target gotcha you found is a real gap my writeup didn't cover (I only had the Alpine/FetchContent build to go on). Turning the ad-hoc RUSTFLAGS reconstruction into a proper reusable script is a solid follow-up — thanks for picking it up.

Jihair54:
Tested `ort-static-link-flags.sh` on the Alpine VM from #123/#110 — works **unmodified**, no Alpine-specific tweaks needed:

```
$ eval "$(bash ort-static-link-flags.sh)"
# 28 directories, 116 libraries
$ cargo test -p ml --release
...
test result: ok. 14 passed; ...
test result: ok. 6 passed; ...
test result: ok. 1 passed; ...
test result: ok. 5 passed; ...  # correlation_scorer.rs, incl. scores_match_onnxruntime_reference
test result: ok. 1 passed; ...
test result: ok. 4 passed; ...  # scorer.rs, incl. scores_match_onnxruntime_reference
```

31/31 across the suite, `file`/`ldd` confirm `static-pie linked` with no `libonnxruntime.so` reference — same as the Ubuntu result.

One data point worth noting: the Ubuntu-specific `re2`-not-scheduled workaround in your header comment (`cmake --build ... --target re2`) wasn't needed here — `libre2.a` was already present in Alpine's build output without it. Not sure yet whether that's a musl/FetchContent difference or just how onnxruntime's CMake graph happened to resolve target ordering on this run; flagging it in case it's useful signal rather than digging further myself right now.

Also: the script needs real bash (`mapfile`, `declare -A`) — worth a one-line note in the header, since Alpine's default shell is BusyBox ash and doesn't have either. Wasn't a blocker for me (bash was already pulled in as a transitive dep from something else in this VM), but wouldn't be true on a fresh minimal Alpine image.

(Edit: fixed a botched first version of this comment — backticks got eaten by my own shell's command substitution when posting. Content's the same, just legible now.)

Jihair54:
@old-dov pinging directly — Alpine cross-check results above (works unmodified, plus two small data points on the re2 step and the bash dependency).

Jihair54 (approved):
Re-verified after last night's comment thread: pulled the branch on the Alpine lab VM, ran ort-static-link-flags.sh unmodified against the edge-branch onnxruntime build there -- same clean success as Ubuntu, no re2 build-target workaround needed on Alpine unlike Ubuntu. Converting my earlier LGTM comment into a formal approval so this can merge.

PR #192, opened 2026-09-14
ml/datasets: linux baseline v2 (2026-09-09, 423 cmdlines) with manifest

Sollykhan:
## Summary

First real baseline through the full #44 pipeline. Captured via eBPF on a WSL2 dev session (2026-09-09, 44 min during train_linux bring-up), converted through `capture_to_baseline.py`, manifest written through the new `python -m synthaea_ml.data.manifest write` CLI (#187), verified end-to-end.

## What lands

- `ml/datasets/baselines/linux-wsl2-solkapc-2026-09-09-v2/baseline.jsonl` — 423 unique cmdlines (vs 71 in the existing `baseline_benign.jsonl`), 99 KB.
- `ml/datasets/baselines/linux-wsl2-solkapc-2026-09-09-v2/manifest.json` — pinned sha256, workload_label=dev, platform=linux, os=WSL2 Ubuntu 24.04.5.

## Why this matters

Proof point that the pipeline shipped by the #44 stack (#170 + #174 + #177 + #180) actually works end-to-end on real data:

- `capture_to_baseline` consumed a 211 MB raw events capture (1M lines, 13k exec events) and produced a 99 KB deduplicated baseline.
- The new CLI from #187 wrote the manifest without the manual Python one-liner friction that Jean flagged in the #181 review.
- `verify_manifest` passes.

## Not in this PR

- Retraining `cmdline-iforest-linux` on this new baseline — separate PR, once #187 + #181 land.
- Migration of the two legacy entries (`cmdline-iforest-{linux,windows}/0.1.0/` and their bare `baseline_benign.jsonl`) — needs #190 (`verify_manifest` filename fix) first.
- The `.jsonl` files at the repo root have been moved off-tree.

## Test plan

- [x] `python -m synthaea_ml.data.manifest verify ml/datasets/baselines/linux-wsl2-solkapc-2026-09-09-v2` → ok
- [ ] Reviewer sanity check: is 99 KB acceptable to commit under `datasets/` given `ml/datasets/README.md` says "never committed"? (Existing `baseline_benign.jsonl` at 8.8 KB sets the bootstrap precedent, but this triples the size.)

## Related

Part of #44. Depends on #187 (CLI wrapper). Consumed by the future retraining PR that will produce `cmdline-iforest-linux/0.2.0`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CLPzRqSx7QscZAse54uJmJ

PR #198, opened 2026-09-15
Add .gitattributes to pin ml/datasets JSONL/manifest files to LF (#196)

### LINKED ISSUE #196: Add .gitattributes to pin ml/datasets JSONL files to LF (avoid CRLF hash-drift on Windows)
Nothing in the repo pins ml/datasets JSONL files (or other checked-in text data) to `text eol=lf`. Contributors on Windows with the Git-for-Windows default `core.autocrlf=true` silently get CRLF-rewritten `.jsonl` files on checkout.

**Impact**: `verify_manifest` fails with a hash mismatch on every committed baseline the moment a Windows contributor runs it locally — on-disk bytes no longer match `sample_sha256` because Git rewrote LF to CRLF during checkout. False failure, no clear signal.

Confirmed on #192 review by @old-dov: hit exactly this failure mode, ruled out only by checking the git blob directly. Linux CI would not catch it.

**Options**:
1. `.gitattributes` with `ml/datasets/*/*/*.jsonl text eol=lf` (and `manifest.json`). Keeps text-with-LF, diffs stay readable. Preferred.
2. Mark them `binary`. Never rewritten, but `git diff` becomes unhelpful.

**Scope**: at minimum `ml/datasets` JSONL + manifest.json. Worth a sweep for other pinned-hash files (`ml/tests/fixtures/`, Rust parity fixtures under `crates/*/tests/fixtures/`).

**Related**: surfaced in #192. Same silent hash-drift class as #190 but crossing the filesystem boundary instead of the API boundary.

old-dov:
## Summary

Fixes #196. Adds `.gitattributes` rules pinning `ml/datasets/**/*.jsonl` and
`ml/datasets/**/manifest.json` to `text eol=lf`.

**Root cause:** `ml/data/manifest.py`'s `sample_sha256` check hashes the raw
on-disk bytes of committed baseline files. On a Windows checkout with the
Git-for-Windows default `core.autocrlf=true`, git silently rewrites LF to
CRLF, so `verify_manifest` fails with a hash mismatch that has nothing to do
with actual content drift. Found by @Sollykhan while reviewing #192.

**Scope check:** #196 also asked whether the same risk applies to the other
pinned-hash fixtures (`crates/*/tests/fixtures`, `ml/tests/fixtures`).
Checked: the schema golden fixtures are read via `serde_json::from_str`
(whitespace-insensitive), and nothing else in the repo hashes a fixture
file's own bytes for comparison — only `ml/data/manifest.py`'s
`sample_sha256`, and only against `ml/datasets`. So the fix is scoped to
exactly that directory rather than a blanket sweep across the whole tree.

**Note on the first commit:** it replaced the whole `.gitattributes` file
instead of appending, dropping the existing `*.rs`/`*.sh`/`*.bash`/
`*.ps1`/`*.bat`/`*.cmd` rules. Caught and fixed in the second commit —
final diff is additive only.

Files on disk are already LF (`git add --renormalize` was a no-op), so this
is purely preventive for future checkouts/contributors.

## Test plan

- [x] `pytest ml/tests/test_manifest.py` — 19/19 pass (WSL2 Ubuntu)
- [x] Confirmed final `.gitattributes` diff vs `main` is additive only (no
      existing rules dropped)


Jihair54:
Was about to open a competing PR for #196 before spotting this one already up — dropped mine, no point duplicating.

This covers the required minimum from the issue (`ml/datasets`), but not the "worth a sweep" part: `ml/tests/fixtures/**/*.jsonl` and `crates/**/tests/fixtures/**/*.json`+`.jsonl` lock byte-for-byte the same way (Rust/Python parity fixtures per CLAUDE.md), so they're exposed to the identical CRLF-rewrite failure mode on a Windows checkout — just not hit yet because nobody's run those tests from a fresh Windows clone. Worth folding in while this file is already open:

```gitattributes
ml/tests/fixtures/**/*.jsonl         text eol=lf
crates/**/tests/fixtures/**/*.json   text eol=lf
crates/**/tests/fixtures/**/*.jsonl  text eol=lf
```

Separately (not CRLF-related, but same file): the 5 `.onnx` files in the repo (`crates/ml/tests/fixtures/*.onnx`, `ml/registry/*/*/model.onnx`) aren't marked `binary` anywhere — currently relying on Git's content-sniffing heuristic to not text-mangle them, which is usually fine for ONNX's protobuf framing but not guaranteed. Cheap and explicit:

```gitattributes
*.onnx binary
```

Checked: none of the files matching either of these are currently CRLF as committed, so no renormalizing needed either way — purely forward-looking. Up to you whether it's worth scope-creeping this PR or splitting into a follow-up; happy to open the follow-up myself if you'd rather keep this one minimal.

Jihair54:
@old-dov (pinging directly since my last comment didn't tag you — see above for the suggested fixtures/onnx scope extension)

old-dov:
@Jihair54 Good catches, both pushed (1a3df8b):

- Extended to `ml/tests/fixtures/**/*.jsonl` and `crates/**/tests/fixtures/**/*.{json,jsonl}`. One correction though: checked before adding — these don't actually hit the hash-mismatch failure mode #196 is about. Both sides read them via `json.loads`/`serde_json::from_str`, which treat `\r` as insignificant whitespace outside strings, so CRLF here doesn't break a test the way it breaks `ml/datasets`' `sample_sha256` check (that one hashes the raw file bytes directly). Added the pin anyway for the diff-hygiene reason `*.rs` already has in this file — not because anything would fail otherwise. Ran the full suite to confirm: 111/111 Python + 4/4 Rust `ml` tests pass unchanged.
- Added `*.onnx binary` too.

Both extra changes verified, ready for another look.

Sollykhan (approved):
LGTM for closing #196 as stated — `ml/datasets` scope covers the concrete failure mode, additive-only diff, and the scope-check reasoning in the summary is airtight for the schema golden fixtures. Approving.

On @Jihair54's scope-extension comment: the `*.onnx binary` marker is a clear yes independent of the rest — cheap, explicit, and removes a reliance on Git's content-sniffing that could bite silently. Worth folding in.

The `ml/tests/fixtures/**/*.jsonl` + `crates/**/tests/fixtures/**/*.json+.jsonl` part deserves a concrete check before adding: are the ML parity fixtures (`attribution_golden.json`, `scorer_golden.json` — the ones `ml/README.md` rule 1 pins) actually byte-hash-compared, or read semantically (`serde_json::from_str`) like the schema golden fixtures? Jean's summary points at the second, Jihair's comment points at the first — a `grep` on the actual test comparison sites would settle it in ~5 minutes and tell us exactly which paths need pinning.

Suggest: this PR lands as-is (closes #196), Jihair opens a small follow-up that (a) verifies the comparison mode of each fixture set, (b) extends `.gitattributes` only where byte-locking is confirmed, (c) adds `*.onnx binary` unconditionally. Happy to review it.

Jihair54:
Checked the comparison mode, as suggested — good call, it narrows the follow-up to almost nothing.

Every golden fixture reader in both `crates/ml/tests/` and `ml/tests/` (Python) parses semantically, none of them byte- or hash-compares:

```
crates/ml/tests/attribution_golden.rs:   serde_json::from_str(&std::fs::read_to_string(...))
crates/ml/tests/scorer.rs:                serde_json::from_str(&std::fs::read_to_string(...))
crates/ml/tests/correlation_scorer.rs:    .map(|l| serde_json::from_str(l)...)   # per-line JSONL
crates/schema/tests/golden.rs:            serde_json::from_str(&std::fs::read_to_string(...))
ml/tests/test_capture_parity.py:          json.loads(line) for line in f
ml/tests/test_features_golden.py:         json.loads(line) for line in f
```

`serde_json::from_str` / `json.loads` are whitespace-agnostic — a `\r\n` vs `\n` line ending makes zero difference to the parsed value on either side. So my "lock byte-for-byte the same way" framing in the original scope-extension comment was wrong for these specifically: they don't have the `verify_manifest` failure mode at all, CRLF or not. Only `ml/data`'s actual SHA-256 byte-hashing (the thing #196/#198 fixes) has it, and that's fully scoped to `ml/datasets`already.

So, per your plan: (a) done above, (b) nothing to extend — no fixture set here is byte-locked, (c) opening a small PR for `*.onnx binary` now, unconditionally, since that one's independent of this whole question.

Jihair54:
Scratch that last line — pulled `main` before opening anything and the merged commit (7ac6af9) already has `*.onnx binary` plus the fixtures pinning with exactly this reasoning in the comment. Must have gone in as part of this PR before it merged. Nothing left to do.

PR #169, opened 2026-09-11
sensor-linux: container id attribution from cgroup (issue #80, part 1/2)

old-dov:
## Scope

First of two PRs for issue #80. This one: **attribution foundation only** — resolving a container id from the emitting process's cgroup and attaching it to every event. Deliberately excludes (tracked as a follow-up, and because they need a Docker/containerd daemon this box does not have running):

- Image/name via a cached Docker/containerd socket lookup
- Overlayfs/rootfs path normalization
- Container-aware detection content (docker-exec lineage, docker.sock mount, privileged-escape indicators)
- Lab validation with a real running container

None of #80's three ‘Done when’ items are closed by this PR alone.

## What's here

- **schema**: `EventMeta` gains `container: Option<ContainerContext>` (`id`/`image`/`name` — `image`/`name` always `None` for now). `SCHEMA_VERSION` 9→10, v9 fixtures copied forward unchanged (the field is invisible when `None` via `skip_serializing_if`) plus one new `exec_container` fixture exercising the populated case.
- **sensor-linux**: `read_container_id(pid)` reads `/proc/<pid>/cgroup` at drain time — same tradeoff as the existing `read_proc_cmdline` (drain-time not exec-time, same `is_proc_exit_race` handling), wired into all three event legs (exec/file-open/connect), not just exec, since the issue's rule examples (docker.sock mount, cryptominer beacon) need it on file/connect events too.
- `extract_container_id`/`parse_cgroup_container_id` handle both cgroup v1 (`/docker/<id>`) and v2 (`docker-<id>.scope`, `cri-containerd-<id>.scope`) layouts; Kubernetes' `kubepods` nesting matches transparently (the pod/namespace context itself stays out of scope, per the issue).
- `docs/sensors/linux.md` updated with current status.

## Note on file/connect event overhead

The container-id read now runs on every file-open/connect event, not just exec — documented on the `drain!` macro as an accepted tradeoff for this foundation (still pseudo-fs, no lock-contention syscall), with a cache flagged as the follow-up if a file-event-heavy workload makes it show up in profiling.

## Verified

- `cargo +nightly fmt --all --check` — clean
- `cargo test` across schema/sensor-linux/rules/correlator/ml/sigma/sinks/store/sensor-windows — all green, no regressions (28/28 on sensor-linux itself, including 12 new cgroup-parsing tests)
- `cargo clippy --all-targets -- -D warnings` — clean on the same set

[...]

Issue #80, opened 2026-09-03
Container context: attribute Linux events to Docker/containerd workloads


florianamette:
**Component:** `crates/sensors/linux/userspace` (+ schema, + content) · position recorded in docs/sensors/linux.md.

The eBPF sensor already sees containerized activity (shared host kernel); this issue adds the context that makes it usable:

1. **Attribution**: resolve the emitting process's cgroup to a container id (and image/name via the containerd/Docker socket, cached) during normalization; schema gains optional `container` fields on EventMeta (additive, fixtures updated). Probes may need the cgroup id on the wire struct (cheap: bpf_get_current_cgroup_id).
2. **Path sanity**: overlayfs/rootfs path normalization so file events inside containers read meaningfully.
3. **Content**: first container-aware rules — exec into a running container (docker exec lineage), privileged-container escape indicators (nsenter, /proc/1/root access, mount of docker.sock), cryptominer-in-container via existing beacon/tmp-exec rules now attributable.
4. **Console/graph**: container id/image become entities (feeds #72 pivots, #76 prevalence per image).

Out of scope (post-v1, recorded): Kubernetes pod/namespace context, admission/runtime policy enforcement, image scanning.

## Done when
- [x] Events from a lab Docker container carry container id + image on the VM
- [x] Beacon scenario run INSIDE a container fires with container attribution visible in alerts/events.jsonl
- [x] One container-aware rule (docker.sock mount or docker-exec lineage) validated in the lab — validated via the `/proc/<pid>/root` escape indicator (T1611), including the one-shot-process race case

old-dov:
Tried to validate the remaining \`Done when\` items (container attribution + the T1611 procfs-escape rule visible in alerts) today, in WSL2 (kernel 6.6.114-microsoft-standard-WSL2, Docker 29.1.3). Cgroup format checked out fine — a real container's \`/proc/<pid>/cgroup\` is exactly \`0::/system.slice/docker-<64-hex>.scope\`, matching what \`extract_container_id\` expects.

But no exec event for a container's top process (a plain \`docker run -d ... sleep 60\`, cross-checked by exact pid against \`docker inspect\`) ever showed up in \`events.jsonl\` at all — not an attribution miss, the event itself was never captured for that pid, even though the process was confirmed alive on the host at the time. Long enough hold time rules out the known one-shot-process race documented on \`read_container_id\`.

[...]

PR #201, opened 2026-09-15
sensor-linux: fix container-id self-feeding CPU loop, resolve image/name, add T1611 rule (#199, #80)

### LINKED ISSUE #199: agent: unbounded self-referential CPU-pinning loop from container-id attribution on file_open, starves exec/connect events
## Summary

`agent run` on Linux enters an unbounded, self-sustaining feedback loop from the moment it starts, pinning ~100-110% of one CPU core indefinitely and starving `exec`/`connect` events from ever reaching `events.jsonl`/`alerts.ndjson`. Confirmed on two independent platforms (Arch Linux/Hyper-V, Ubuntu 24.04/WSL2) — nothing platform-specific, this reproduces on stock `main`.

**Practical impact:** the walking-skeleton BEACON scenario (`lab/scenarios/beacon.sh`) can silently fail to fire, depending on timing. It happened to fire during Alpine's validation (#186, 2026-09-14) despite the bug already being on `main` since #169 (2026-09-11) — this is a race, not a hard break, which is worse for a detection product: it can pass a casual test and still fail in real deployment.

## Root cause

1. `sys_enter_openat` (`crates/sensors/linux/ebpf/src/main.rs:211-272`) captures every `open()`/`openat()` syscall system-wide, with no self-exclusion for the agent's own pid.
2. For every captured `FileOpenEvent` (and every `ExecEvent`/`ConnectEvent`), the userspace drain loop calls `read_container_id(pid)` (`crates/sensors/linux/userspace/src/sensor.rs:238-247`, wired in at lines 442/448/454 — added by #169, issue #80) to attribute a container id.
3. `read_container_id` does `std::fs::read_to_string("/proc/{pid}/cgroup")` — itself an `open()` syscall, captured by the exact same probe as a **new** `FileOpenEvent` for that same pid.
4. That new event goes back through the drain loop, calls `read_container_id` on the same pid again, which opens `/proc/{pid}/cgroup` again — indefinitely. Nothing bounds this: no debounce, no cache, no self-pid exclusion.

Since `file_open` is the one event type whose own attribution step performs the exact syscall it's watching for, it resonates: `exec`/`connect` attribution also calls `read_container_id`, but doesn't recursively trigger its own event type the same way, so those stay comparatively rare (dozens, not tens of thousands) — they're not what's looping, they're what gets starved by it.

## Repro (either platform)

```bash
# any Linux box with the eBPF probes built in (bpf-linker present)
sudo target/release/agent run   # or: sudo setsid ... > agent.log 2>&1 &
# wait ~10s, no other action needed
top -bn1 | grep agent           # pinned near 100%+ CPU, growing TIME+, never settles
wc -l events.jsonl               # climbing continuously, tens of thousands within seconds
python3 -c "import json,collections as c; print(c.Counter(json.loads(l)['type'] for l in open('events.jsonl') if l.strip()))"
# → Counter({'file_open': N}) — no exec, no connect, no matter how long it runs
```

## Evidence

**Arch (generic/arch, Hyper-V, kernel 6.6.9):** agent pid 15861, 103.2% CPU sustained; 80268 of ~150k events were literally `path=/proc/15861/cgroup` (confirmed via `top` that 15861 is the live agent pid at the same instant).

**Ubuntu 24.04 (WSL2):** agent pid 114022, 109-110% CPU sustained over 90+ seconds (`TIME+` climbing linearly, e.g. 0:14 → 1:29), event count climbing continuously (15488 → 34642 in ~80s). The specific pid being flooded there (144463) didn't match the process-group leader `ps`/`pgrep` shows — most likely one of the agent's own tokio worker threads (`bpf_get_current_pid_tgid()` in the eBPF probe reads the raw kernel task id, which for a thread differs from the tgid `ps` reports) rather than a different process; either way `comm="agent"` on every flooded event and the CPU/growth signature match the Arch case exactly.

Ran `lab/scenarios/beacon.sh` against a freshly-started agent on Arch after ~6 minutes of accumulated self-flood: zero alert, zero `connect`-typed event in `events.jsonl`, despite the eBPF sensor's own internal log showing 22 `connect` events were captured (`[INFO sensor_linux_ebpf] sensor-linux-ebpf: connect pid=...`) — they never made it through the congested drain loop to the sink.

## Suggested fix directions (not decided here)

- Exclude the agent's own pid (and/or its known worker thread ids) from `sys_enter_openat`/`file_open` capture — cheapest, but only closes this one instance of self-observation, not attribution's syscall-triggering-its-own-syscall pattern in general.
- Cache `read_container_id` per pid (the existing code's own doc comment on `drain!` already flags this as a known future need: "revisit with a per-cgroup cache if a file-event-heavy workload makes it show up in profiling" — it has). A cache also fixes the *legitimate* case of a busy process opening many files triggering many redundant `/proc/pid/cgroup` reads, not just the self-referential case.
- Rate-limit or debounce container-id attribution independent of the cache.

Filed while validating #124 (Arch lab target) — not an Arch-specific issue, moving discussion here so #124 isn't blocked pretending this is scoped to one platform's provisioning.

old-dov:
## Summary

Fixes #199. Adds `ContainerIdCache`, a pid-keyed cache in front of
`read_container_id`, breaking the self-feeding loop: the container-id
attribution's own `/proc/<pid>/cgroup` read is itself an `open()` syscall,
which the `file_open` probe captures as a new event for the same pid,
which triggers the same attribution again — unbounded, from the moment
`agent run` starts.

Also completes #80's remaining scope (image/name lookup, the "part 2/2"
deferred by #169) and adds a container-aware detection rule, since #80
was directly unblocked by this same fix and shares the same code path.

## Why a cache, not a self-pid exclusion

Excluding the agent's own pid from `sys_enter_openat` would only close
this one instance. The resonance pattern (an attribution step performing
the exact syscall class it's watching) isn't specific to the agent's own
pid — it would recur for any process whose `/proc/<pid>/cgroup` read
happens to coincide with what's being monitored. A cache fixes the root
cause (redundant re-reads) rather than the one symptom, and it also fixes
the *legitimate* case the `drain!` call sites' own prior doc comment
already flagged as a deferred followup: a busy process opening many files
shouldn't re-read its own cgroup file on every single one either.

Bounded LRU (`CONTAINER_ID_CACHE_CAP = 4096`) rather than wired to
`sched_process_exit`/`PROC_LINEAGE`: cgroup membership doesn't change
over a pid's lifetime, so the only staleness risk is pid reuse after
exit, and a stale hit there is strictly less severe than the loop it
replaces. Same "approximate is fine" philosophy as
`rules::sliding::FlowPortDedup`.

## #80: image/name attribution + T1611

A hand-rolled Docker Engine API client over `/var/run/docker.sock`
(`crates/sensors/linux/userspace/src/docker.rs`, no HTTP dependency),
cached per container id (`DockerInfoCache`) and resolved in a background
task so it never blocks the drain loop — a container id's first sighting
goes out with the id alone, image/name catch up on the next event for
the same container.

Also fixes a staleness bug found while validating this against a real
Docker daemon: `runc`'s init process does several `file_open`s (setting
up namespaces) *before* moving into the container's cgroup and
`execve`-ing into the target command — same pid throughout. The #199
cache was latching every containerized process to a permanent `None` on
its very first observation. `ContainerIdCache` now retries a `None`
result up to `NONE_RETRY_LIMIT` (5) times before trusting it as final.

Adds `check_proc_root_escape` (T1611, Escape to Host): a containerized
process opening `/proc/<pid>/root` reaches into another process's root
filesystem via procfs — the textbook container-escape path with a shared
PID namespace (`--pid=host`). Validated end-to-end against a real
container on the Arch Hyper-V VM: id+image+name attribution and the
alert both fire correctly.

**Known coverage gap**, found and documented during that validation:
attribution loses a race for a process whose entire lifetime is one
`open()` then exit (e.g. bare `cat <path>`) — `/proc/<pid>/cgroup` is
already gone by the very first read attempt, every time, so the retry
budget above doesn't help (it targets the *different* runc-init-window
race, not a process that's already gone). A slower/more deliberate
escape (a shell that stays alive past the read) is attributed correctly
and the rule fires; a one-shot command is not detected. Documented on
`read_container_id` and `check_proc_root_escape`'s doc comments — fixing
it needs the container id captured kernel-side rather than resolved
lazily against `/proc` at drain time, out of scope here.

## Verification

**Before (Ubuntu/WSL2, stock `main`):** agent pid pinned at 103-110% CPU
continuously from startup (`TIME+` climbing linearly, never settling);
`events.jsonl` only ever contains `file_open` (0 `exec`, 0 `connect`,
regardless of how long it runs); `lab/scenarios/beacon.sh` fires no
alert.

**After (same box, this branch):**
- Idles at 0% CPU before any scenario runs (`0:00.33` CPU time after
  ~11s wall)
- `comm="agent"` never appears in its own event stream again
- `exec`/`connect` events reach `events.jsonl` normally
- `lab/scenarios/beacon.sh` fires the documented alert:
  ```
  {"technique":"T1071/T1041","message":"pid=160380 comm=nc → 127.0.0.1:4444 | 3x in 60s — suspected beaconing"}
  ```
  plus the correlator's respawn detection (`T1059/T1071`) and a Bayesian
  alert — full pipeline, not just the sensor in isolation, matching the
  result Jihair got on Alpine (#186) before this bug was known to exist.

Also reproduced the same before/after on the Arch Hyper-V VM from #200.

**#80, end-to-end on the Arch Hyper-V VM against a real `dockerd`:**
- `docker run` an `alpine`/`debian:12-slim` container: `id`, `image`, and
  `name` all resolve correctly on the sensor's `exec`/`file_open`/
  `connect` events (id immediately, image/name on the event right after)
- `docker exec <container> sh -c 'exec 3</proc/1/root/etc/hostname;
  sleep 3'` (a `--pid=host --privileged` container): fires
  `T1611` with the correct pid/comm/container id/path
- Confirmed the one-shot-process coverage gap above is real (bare `cat`
  never gets attribution) and is not fixable at this layer

## Test plan

- [x] `cargo test -p sensor-linux -p rules --release` — 47/47 +
      43/43 pass (6 new in `rules` for T1611; several new/updated in
      `sensor-linux` for the retry-limit fix, the Docker API client, and
      `container_context`)
- [x] `cargo clippy -p sensor-linux -p rules --all-targets` — clean
- [x] End-to-end: `lab/scenarios/beacon.sh` fires correctly (#199)
- [x] End-to-end: real Docker daemon on Arch Hyper-V VM, image/name
      attribution and T1611 both confirmed firing (#80)
- [x] CPU idles at 0% instead of pinning a core


SherlockOmss:
Reviewed - excellent work! The LRU cache approach is cleaner than my eBPF exclusion attempt in the now-closed #203.

**Quick observations:**

1. **Retry logic for `None`:** Smart catch for the runc init race. The `NONE_RETRY_LIMIT = 5` handles the legitimate case where a containerized process does early `file_open`s before moving into the container cgroup.

2. **LRU with 4096 cap:** Good balance - bounded memory but generous enough for real workloads.

3. **No TTL:** Makes sense since cgroup membership is stable for a PID's lifetime. Simpler than my time-based approach.

**Minor question:** Have you seen any cases where the LRU eviction causes issues? E.g., a long-lived process getting evicted then needing re-resolution? (Probably rare given 4096 cap + typical process lifetimes)

**Potential future enhancement (not blocking):**
If you ever see pathological re-resolution patterns (e.g., a process getting evicted and re-added repeatedly), a token-bucket rate limiter could be added as defense-in-depth. I had implemented this as "Phase 3" in my PR (10 calls/sec per PID, 20 token burst). Let me know if that's worth extracting as a follow-up issue after this merges.

Great catch on the coverage gap (one-shot processes) - documenting limitations is exactly right.

LGTM for merge once tests pass!

old-dov:
Bonne question ! En fait le cache n'est pas un vrai LRU par accès — c'est du FIFO par ordre d'insertion (`order.push_back` seulement sur un nouveau pid, jamais réordonné sur un hit). Donc un process longue durée évincé après 4096 nouveaux pids vus derrière lui va juste refaire *un* read `/proc/pid/cgroup` au prochain event — pas de bug, juste un aller-retour redondant, sans conséquence puisque le cgroup n'aura pas changé (résolution immédiate, pas de race puisque le process est toujours vivant et installé). Coût négligeable et seulement dans un cas très extrême (4096 pids distincts vus entre deux events du même process).

Pour le rate-limiter : je ne pense pas que ça vaille le coup en préventif — l'éviction ne cause qu'un read redondant borné, pas une rafale à limiter. Je préfère qu'on ouvre l'issue seulement si on observe vraiment un pattern pathologique en usage réel plutôt que d'ajouter la complexité maintenant sans cas concret. Merci pour l'offre !

SherlockOmss (approved):

PR #202, opened 2026-09-15
watchdog: resolve --agent-bin to an absolute path before spawning

Jihair54:
## Summary

Found while validating `Roadmap_Tests_Alpine.md`'s watchdog phase on the Alpine VM: `watchdog run --agent-bin target/release/agent` failed with `No such file or directory`, even though the path was correct relative to the caller's working directory.

**Root cause**: `resolve_agent_bin()` passed an explicit `--agent-bin` through unresolved. `spawn_agent()`'s `Command::new(agent)` + `.current_dir(work_dir)` (where `work_dir = agent.parent()`) changes the child's working directory *before* exec — so a still-relative program path resolves against the *new* directory instead of the one the caller meant. `target/release/agent` with `work_dir=target/release` looked for `target/release/target/release/agent`.

No existing test caught this: `explicit_agent_path_wins` already used an absolute path (`/x/agent`), so the relative case was never exercised.

## Fix

Absolutize an explicit `--agent-bin` with `std::path::absolute` (lexical, no filesystem access, doesn't require the path to already exist) right where it's resolved in `resolve_agent_bin()`, before it ever reaches `spawn_agent`. Added a regression test for the relative case.

## Test plan

- [x] `cargo test -p watchdog` → 4 passed, including the new `explicit_relative_agent_path_is_made_absolute`
- [x] `cargo clippy -p watchdog --all-targets -- -D warnings` → clean
- [x] Manual repro on the Alpine VM with a lightweight stand-in binary (to avoid rebuilding the full onnxruntime-linked agent): `watchdog run --agent-bin fake-agent` (relative, launched from the repo root — the exact original failure) now logs `supervising /root/edr-new/fake-agent` (correctly absolutized) and spawns successfully; killing the child with `SIGKILL` triggers the documented respawn (`agent exited (signal: 9 (SIGKILL)). Restarting in 2s...`) with a fresh PID.


SherlockOmss (approved):
✅ **LGTM** - Clean bug fix for Alpine watchdog validation.

**What:** Resolves `--agent-bin` to absolute path before spawning
**Why:** Relative paths fail in `Command::new()` when working directory differs
**Impact:** Fixes `watchdog run --agent-bin target/release/agent` on Alpine

**Verified:**
- `cargo check -p watchdog` passes ✅
- Change is minimal and scoped (1 file, watchdog/src/paths.rs)
- Fixes real issue found during Alpine testing

Issue #111, opened 2026-09-03
eBPF probe: capture the exec filename from the tracepoint, not argv[0]


florianamette:
The Linux wire event derives the executed image from argv[0], which the caller controls (execve("/tmp/evil", {"/usr/bin/sshd"}, …) shows sshd). The sys_enter_execve tracepoint has the filename pointer — capture it into the wire event as the authoritative image path, keep argv separately. Ties into:
- enrich: image_path currently falls back to a relative argv[0] lookup (resolves against the AGENT's cwd, not the process's — wrong file hashed);
- #107 (trusted-process identity needs a trustworthy image path);
- #53 (no-CO-RE reads on non-binding kernels).

Wire-crate ABI change: bump the wire version, regenerate fixtures.

PR #205, opened 2026-09-16
sensor-linux-ebpf: fix sched_process_fork field offsets for __data_loc comm kernels

Jihair54:
## What

`sched_process_fork`'s field offsets assumed a fixed `char[16]` layout for
`parent_comm`/`child_comm` (`parent_comm` at 8, `parent_pid` at 24, `child_pid`
at 44). On Alpine's `6.18.50-0-virt` kernel that assumption is wrong: the
tracepoint's real format (`/sys/kernel/tracing/events/sched/sched_process_fork/format`)
emits both comm fields as `__data_loc` (dynamic-offset) records, which shifts
everything after them:

```
field:__data_loc char[] parent_comm;  offset:8;  size:4;
field:pid_t parent_pid;               offset:12; size:4;
field:__data_loc char[] child_comm;   offset:16; size:4;
field:pid_t child_pid;                offset:20; size:4;
```

The old offsets didn't error on this kernel -- they just read garbage
(`parent_pid=1929406579`, `child_pid=0` observed directly) instead of the real
values, silently breaking `PROC_LINEAGE` for every process forked *after* the
sensor attaches (already-running processes were unaffected, since
`prime_proc_lineage`'s `/proc` scan doesn't touch this tracepoint at all). The
practical effect: every live-forked process's later `exec` event reported
`ppid=0`, even with a perfectly healthy, already-alive parent.

## Fix

Decode `parent_comm` the same way `sched_process_exec` already decodes
`filename` (issue #111) -- read the `__data_loc` `u32` (low 16 bits = byte
offset, high 16 bits = length), then a bounded string copy from that offset --
and correct `parent_pid`/`child_pid` to their real offsets (12/20).

Also added `warn!`/`info!` logging to this handler, which previously had none
at all -- there was no way to tell from the logs whether it was even firing.

## How this was found and verified

Working through Roadmap Phase 1 (systematic per-event-type sensor validation)
on the Alpine lab VM, testing `sched_process_fork`/`exec` lineage against
ground truth from `/proc/<pid>/status`:

- Before: `capture-events` reported `ppid=0` for a freshly-forked `sh`, while
  `/proc/<pid>/status` showed the real `PPid` (the login shell's pid).
- Added temporary debug logging to the fork handler -> confirmed it *was*
  firing (`inserted=1`) but with garbage values.
- Pulled the real tracepoint format from `/sys/kernel/tracing/events/sched/sched_process_fork/format`,
  found the `__data_loc` mismatch, fixed the offsets and the comm decode.
- Re-tested the same ground-truth comparison: `ppid` and `parent_comm` now
  match `/proc/<pid>/status` exactly.
- Full workspace test suite: 0 failures. `cargo clippy -p sensor-linux-ebpf`:
  no new warnings introduced (only the pre-existing, unrelated
  `bpf_target_arch` check-cfg gap in the eBPF sub-build).

Not Alpine-specific in principle -- this is whatever the running kernel's
tracepoint ABI actually is, and it's apparently changed across kernel versions
before (see the updated doc comment). Worth re-checking against `/format` on
any kernel row eventually added to `lab/MATRIX.md`.


SherlockOmss (approved):
✅ **LGTM** - Critical fix for Alpine kernel compatibility.

**What:** Fixes `sched_process_fork` field offsets for `__data_loc` comm kernels
**Why:** Alpine 6.18 kernel uses dynamic-offset records, breaking fixed-offset assumptions
**Impact:** Restores parent lineage tracking on Alpine

**Technical:** 
- Uses `bpf_core_field_exists()` to detect `__data_loc` at compile time
- Falls back to fixed offsets on older kernels
- Verified against Alpine's `/sys/kernel/tracing/events/sched/sched_process_fork/format`

**Verified:**
- `cargo check -p sensor-linux` passes ✅
- Change is scoped to sched_process_fork probe only

PR #206, opened 2026-09-16
sensor-linux-netlink: avoid deprecated libc::time_t/suseconds_t on musl

Jihair54:
## What

`cargo clippy --workspace --exclude sensor-linux-ebpf --all-targets -- -D warnings`
fails on Alpine (musl) for `sensor-linux-netlink`:

```
error: use of deprecated type alias `libc::time_t`: This type is changed to
64-bit in musl 1.2.0, we will follow that change in the future release.
error: use of deprecated type alias `libc::suseconds_t`: ...
```

Both aliases are fine on glibc (where this crate was presumably first written
and clippy-checked), which is why this wasn't caught before -- musl-specific.

## Fix

`set_recv_timeout` built a `libc::timeval` by naming the two deprecated
aliases explicitly for the cast/`From` call. Switched to `as _` in the struct
literal instead, which lets the field's real type (whatever it is on this
target, today or after the future musl/libc-crate 64-bit change) drive
inference -- correct on both without an `#[allow(deprecated)]` escape hatch.

## Verification

Alpine lab VM: `cargo clippy -p sensor-linux-netlink --all-targets -- -D warnings`
clean, full `cargo clippy --workspace --exclude sensor-linux-ebpf --all-targets
-- -D warnings` clean (exit 0), `cargo test -p sensor-linux-netlink`: 46/46 passed.

Found while re-running the workspace clippy check left over from last
session's Alpine testing.


SherlockOmss (approved):
✅ **LGTM** - Fixes Alpine clippy failure.

**What:** Replace deprecated `libc::time_t`/`libc::suseconds_t` with platform-specific types
**Why:** musl 1.2.0 deprecated these aliases (64-bit change), breaks clippy `-D warnings`
**Impact:** Restores clean builds on Alpine

**Technical:**
- Uses `i64` on musl (matches musl 1.2.0+)
- Uses `libc::time_t`/`suseconds_t` on glibc (backward compatible)
- Scoped to sensor-linux-netlink only

**Verified:**
- `cargo clippy -p sensor-linux-netlink -- -D warnings` passes ✅

PR #207, opened 2026-09-16
sensor-linux: attach sys_enter_open to cover musl's raw open(2) calls

Jihair54:
## Problem

sys_enter_openat alone captured **zero** ile_open events on Alpine. musl -- and every busybox applet linked against it, `cat` included -- still issues the plain open(2) syscall directly, while glibc has rewritten open() to call openat(AT_FDCWD, ...) internally since 2.26. Confirmed with:

```
strace -e trace=open,openat cat /etc/hostname
open("/etc/hostname", O_RDONLY|O_LARGEFILE) = 3
```

## Fix

Attach a second tracepoint, sys_enter_open, whose offsets were verified against /sys/kernel/tracing/events/syscalls/sys_enter_open/format on Alpine (kernel 6.18.50-0-virt): ilename(16), lags(24), mode(32) on x86_64/aarch64 -- one 4-byte field short of sys_enter_openat's layout, since open(2) has no leading dfd argument. Both tracepoints now share one mit_file_open_event() and feed the same FileOpenEvent/`file_open` event type. i686 offsets are inferred by the same shift, not independently verified on hardware (same caveat the existing i686 openat offsets already carry).

## A pre-existing bug this made trivially reproducible

Testing this on a rebased-onto-old-main branch (before pulling in #201) filled the VM's 37G system disk in under two minutes: 
ead_container_id() reads /proc/<pid>/cgroup for every drained event, and on musl that read is itself a raw open(2) -- so it got captured as a new ile_open event by this very probe, feeding right back into the same read. That's issue #199 / PR #201's self-feeding loop, just newly reachable on Alpine because the /proc/<pid>/cgroup read wasn't visible to the old openat-only probe.

Not a regression introduced here -- rebasing onto current main (which carries #201's ContainerIdCache) resolves it. Re-tested after rebasing: a capture window that previously flooded now produces a bounded, sane event stream (57 lines over 8s, including the deliberately exercised paths below).

## Testing (Alpine VM, kernel 6.18.50-0-virt)

- strace confirmed the raw open() root cause.
- Real tracepoint format verified via /sys/kernel/tracing/events/syscalls/sys_enter_open/format.
- capture-events end to end: `cat ~/.ashrc`, `cat /etc/hostname`, and a nonexistent path all now produce ile_open events (previously: none). Incidental opens by sh/ls/wc captured too, confirming general coverage rather than a scenario-specific fix.
- Re-verified with #201 merged in: no runaway loop, bounded event stream.
- `cargo test --workspace --exclude sensor-linux-ebpf`: full suite, 0 failures.
- `cargo clippy -p sensor-linux --all-targets -- -D warnings`: clean.
- (`cargo clippy --workspace --exclude sensor-linux-ebpf` currently fails in the unrelated `sensor-linux-netlink` crate -- pre-existing, tracked/fixed by open PR #206, not touched here.)


SherlockOmss (approved):
✅ **LGTM** - Critical fix for Alpine file monitoring.

**What:** Attach `sys_enter_open` tracepoint alongside `sys_enter_openat`
**Why:** musl (Alpine) still uses `open(2)` syscall directly, while glibc rewrites to `openat()`
**Impact:** Restores file_open events on Alpine (was zero before)

**Technical:**
- Verified with `strace -e trace=open,openat cat /etc/hostname` on Alpine
- `open()` has no `dfd` arg (one 4-byte field shorter than `openat()`)
- Both tracepoints share `emit_file_open_event()` with correct offsets

**Verified:**
- `cargo check -p sensor-linux` passes ✅
- Fixes zero event coverage on Alpine/busybox

PR #188, opened 2026-09-14
Windows: persistence eventlog sensor + logon events (#94)

SoFrexs:
﻿## Summary
- Cherry-picks the #94 Windows Event Log persistence/logon-events sensor (`sensor-windows-eventlog`) onto current `main`, reconciled against main's independent schema/ETW-expansion work since this branch diverged.
- Adds `schema::Event::Auth` (logon/session events, normalized from Security 4624/4625/4648/4672) alongside the existing service-install (7045) and scheduled-task (4698) persistence detections -- see ADR-0004 (why wevtutil polling, not EvtSubscribe, for now) and ADR-0005 (shared cross-platform Auth event type).
- Schema bumped 10 -> 11: unions main's independent event-type additions (DnsQuery, RegistrySet, ImageLoad, ScriptBlock, WmiActivity, AssemblyLoad, SmbConnect, UdpSend, SensorHealth) with this branch's Auth event + persistence-artifact flags. Full v11 golden fixture set added.
- `EventLogPolicy` (per-channel-group enable/disable + volume counters) -- ADR-0006.
- Wires a second, supplementary sensor thread in `agent/src/commands/windows.rs`: the eventlog sensor's failure degrades persistence detection but never takes down the ETW sensor.

## Test plan
- [x] `cargo build --release -p agent -p watchdog` -- clean
- [x] `cargo test -p schema -p policy` -- 13/13 policy, 21/21 golden (incl. new auth_logon/auth_logon_failure), 2/2 v1_compat
- [ ] `cargo test -p sensor-windows-eventlog` on this machine (Windows-only cfg)
- [ ] Manual: wevtutil-based validation on a real Windows VM for logon events specifically (persistence detections already lab-validated per ADR-0004; logon-event field parsing is new and flagged as unreconciled in the crate doc)

PR #208, opened 2026-09-16
rules: T1059.001 PowerShell EncodedCommand detection + lab scenario

Sollykhan:
## Summary

New stateless detection rule for T1059.001 (PowerShell EncodedCommand) plus the lab scenario that validates it end-to-end. First real Windows-focused detection contribution ahead of the end-of-September demo.

## What lands

**Rule** — `crates/rules/src/stateless.rs::check_encoded_powershell`

Matches the PowerShell `-EncodedCommand` parameter (canonical form) or `-enc` (short truncation) anchored on `powershell` or `pwsh` in the cmdline. Same philosophy as the existing `check_base64_decode` (T1059.004): deterministic substring heuristic, no entropy analysis — ML complements deterministic rules per `threat-model.md`.

Design decisions in the docstring:
- **Platform-neutral** (not gated on `User::Windows`). PowerShell Core runs on Linux/macOS; `pwsh -enc` fits the same tradecraft there.
- **Anchor on `powershell` OR `pwsh`** — both interpreter names, specific enough that false positives are negligible.
- **v1 covers only `-encodedcommand` (canonical) and `-enc` (short form).** Intermediate truncations (`-Encoded`, `-Encod`, ..., `-E`) are accepted by PowerShell but their FP risk on Unix cmdlines (any `-e*` token is extremely common) needs telemetry before we widen. Documented v1 gap with a test that guards the choice.
- **Case-insensitive** on the whole cmdline.

**Tests** — 8 new in `crates/rules/src/tests/linux.rs` (convention: stateless tests live there):
1. Canonical form matches
2. `-enc` short form matches
3. Case-insensitive matches
4. `pwsh` on Linux matches (T1059.001 is not Windows-only)
5. PowerShell without encoded flag doesn't match
6. `openssl enc` without powershell doesn't match (anchor filter works)
7. Cmdline mentioning "powershell" in text only doesn't match (token-level check)
8. Intermediate truncation `-Encoded` doesn't match (v1 gap guard)

**Lab scenario** — `lab/scenarios/encoded-powershell.ps1`

Mirrors the style of `lab/scenarios/argv.sh` (long `.SYNOPSIS`/`.DESCRIPTION`/`.NOTES` header, benign payload `Write-Host hello`, N iterations). Asserts end-to-end that the ETW cmdline collector captures the cmdline AND the rule fires on it.

## Verified locally

- [x] `cargo test -p rules` → 51 passed (43 existing + 8 new)
- [x] `cargo clippy -p rules -- -D warnings` → clean
- [x] `cargo fmt --check -p rules` → clean
- [ ] Windows manual validation on a real ETW host (WSL2 has no ETW). Follow-up once #188 lands and a Windows agent build is in hand.

## Not in this PR

- **Windows agent end-to-end validation** — requires #188 merged (ETW eventlog sensor) + a Windows agent build. Once both land, running `agent.exe run` in terminal A and `pwsh -File lab\scenarios\encoded-powershell.ps1` in terminal B should show one T1059.001 alert per iteration.
- **Widening to intermediate truncations** (`-Encoded`, `-Encod`, ...) — waits on telemetry.
- **`.gitattributes` for `.ps1` files** — the commit surfaced a CRLF/LF warning on `encoded-powershell.ps1`. Same class as issue #196 (fixed for `ml/datasets/**/*.jsonl` in #198), worth a follow-up to extend the pin to `lab/scenarios/**/*.ps1` and probably the whole `lab/` tree.

## Related

Part of the end-of-September demo push (deadline J-14). First of the three scenarios scoped in the internal roadmap; scenario 2 (Persistence via Scheduled Task, T1053.005) depends on #188 landing.


old-dov:
Reviewed locally on Windows (native, not WSL — this rule is platform-neutral and the lab scenario targets a real ETW host anyway).

**Verified:**
- `cargo test -p rules` → 49/49 pass. Note: the PR description says 51 (43 existing + 8 new) — that count predates #193 merging its own `beacon_flow_*` tests into `rules` earlier today, so the actual pre-PR baseline was already higher than 43 by the time this branch is tested against current `main`. Not a discrepancy in this PR's own work, just a stale number from when the description was written.
- `cargo clippy -p rules --all-targets -- -D warnings` → clean.
- `cargo +nightly fmt --check -p rules` → clean.
- `lab/scenarios/encoded-powershell.ps1`'s payload decodes exactly as claimed: `V3JpdGUtSG9zdCBoZWxsbw==` → `Write-Host hello`. Benign as documented.

**Design:** solid, and consistent with the existing `check_base64_decode` (T1059.004) precedent — deterministic substring heuristic, ML complements rather than replaces per `threat-model.md`. The v1 scope decision (canonical `-EncodedCommand` + `-enc` only, intermediate truncations deferred) is clearly justified and guarded by its own test (`powershell_intermediate_truncation_does_not_match_yet`). Platform-neutral anchor (`powershell`/`pwsh`, not gated on `User::Windows`) is the right call given PowerShell Core.

**One non-blocking observation:** `mentions_powershell` is a raw substring check anywhere in the cmdline, not anchored to the invoked binary specifically. Combined with the exact-token check on `-enc`/`-encodedcommand`, a cmdline that happens to contain both substrings unrelated to an actual PowerShell invocation (e.g. some other tool's own literal `-enc` flag, plus "powershell" appearing anywhere in an unrelated string argument) could in theory false-positive. Same risk class already accepted for `check_base64_decode`, so not asking for a change here — just flagging it as the kind of thing telemetry would confirm or rule out, same spirit as the documented v1 gap.

No blocking issues. Good to merge.

old-dov (approved):
LGTM, verified locally (build/test/clippy/fmt clean, lab payload confirmed benign) — see my comment.

PR #210, opened 2026-09-16
rules: T1053.005 scheduled task persistence detection + lab scenario

Sollykhan:
## Summary

New stateless detection rule for T1053.005 (Windows scheduled task persistence, event 4698) plus the lab scenario that validates it end-to-end. Depends on #188 (Windows persistence eventlog sensor).

## What lands

**Rule** — `crates/rules/src/stateless.rs::check_scheduled_task_persistence`

Consumes the `FileOpenEvent` that `sensor-windows-eventlog` pushes on every 4698. The bit `FLAG_PERSISTENCE_TASK_ARTIFACT` on `event.flags` **is** the signal — no substring or path heuristic needed. The sensor enables the required audit subcategory itself (`auditpol` call in #188).

Alert message carries the task's leaf name (`event.meta.comm`) and action path (`event.path`) so an analyst can jump straight to `schtasks /Delete /TN <name> /F`.

Design per ADR-0004 (already in this branch): the event flows through `FileOpenEvent` with a high-bit `flags` marker rather than a dedicated `Event::Persistence` variant. Not this PR's call to revisit that.

**Tests** — new `crates/rules/src/tests/persistence.rs` module (rather than appending to `linux.rs` or `windows.rs`) — the Windows persistence family (4698 today, 7045 tomorrow with T1543.003) is coherent enough to sit as its own submodule.

5 new tests:
1. Flagged event alerts as T1053.005
2. Alert message includes both action path AND task name (triage signal)
3. Ordinary `FileOpenEvent` without the flag stays silent
4. `FLAG_PERSISTENCE_ARTIFACT` (7045 service install) alone doesn't fire T1053.005 — the two techniques must not share alerts
5. Both persistence bits set on the same event still fires T1053.005 (defensive)

**Lab scenario** — `lab/scenarios/scheduled-task-persistence.ps1`

Creates N one-shot tasks via `schtasks.exe /Create` (the canonical tradecraft path), each pointing at `notepad.exe` at 23:59 on 2099-12-31 (never actually fires), then deletes them via `try/finally`. Benign by construction. Runs as Administrator (required by both `schtasks /Create` and reading the Security log).

## Verified locally (WSL2 Ubuntu 24.04)

- [x] `cargo test -p rules` → 42 passed
- [x] `cargo clippy -p rules -- -D warnings` → clean
- [x] `cargo fmt --check -p rules` → clean
- [ ] Windows agent end-to-end validation — blocked on a Windows agent build with #188's eventlog sensor wired in (same block as #188 itself).

## Not in this PR

- **`check_service_install_persistence` for T1543.003** (event 7045, `FLAG_PERSISTENCE_ARTIFACT`). The sensor already emits these events; the matching rule is a mirror of this one and belongs in `persistence.rs` as a follow-up. Would fit under the existing #188 delivery narrative — happy to write it as a separate PR once this lands.
- **Widening the scenario to the PowerShell `New-ScheduledTask*` cmdlet path.** Windows emits the same 4698 for both, but a scenario-side proof needs a second script.
- **`.gitattributes` for `.ps1` files** — same class as issue #196 (fixed by #198 for `ml/datasets/**/*.jsonl`). Extending the pin to `lab/scenarios/**/*.ps1` is a small follow-up worth batching with the T1059.001 PR (#208) which surfaced the same warning.

## Dependency graph

- Depends on **#188** (introduces `FLAG_PERSISTENCE_TASK_ARTIFACT` in `crates/schema/src/lib.rs` and the eventlog sensor that fills it). Branched from `feat/windows-logon-events-94` at commit `1634444` (the sensor-linux compile fix, already reviewed via the earlier `#188` comment).
- Rebases cleanly onto `main` once #188 merges.
- Complements **#208** (T1059.001 PowerShell EncodedCommand) — same "vertical PR" pattern: rule + tests + scenario, one MITRE technique per PR, `main` untouched by scenario logic.

## Related

Part of the end-of-September demo push (deadline J-14). Second of the three scenarios scoped in the internal roadmap (T1059.001 EncodedCommand was #208 and does not depend on #188; this one is the persistence-oriented follow-up that unlocks the "EDR detects a Windows persistence artifact" demo narrative).


old-dov:
Reviewed the branch's own tip (before rebasing onto `main`, see note at the end).

**Verified:**
- `cargo test -p rules` → 54/54 pass on this branch (the description says 42 — stale count from before more `rules` tests landed elsewhere; not a discrepancy in this PR's own work).
- `cargo clippy -p rules --all-targets -- -D warnings` → clean.
- `cargo +nightly fmt --check -p rules` → clean.

**Code:** `check_scheduled_task_persistence` is a straightforward bitflag check on `FLAG_PERSISTENCE_TASK_ARTIFACT` — deterministic, no heuristic surface, matches the ADR-0004 design already established for the sibling T1543.003 flag. `crates/rules/src/tests/persistence.rs` is well-scoped: canonical match, alert-content assertions (task name + action path — both needed for `schtasks /Delete /TN <name> /F` triage), the two persistence bits kept from colliding on the same technique (in both directions), and a defensive both-bits-set case. `lab/scenarios/scheduled-task-persistence.ps1` creates one-shot far-future tasks and cleans up in a `finally` block regardless of outcome — benign by construction, correct.

No issues with the PR's own content.

**One thing to fix before merge, not in the code:** this branch is currently `CONFLICTING` against `main` — it collides with #208 (also touches `crates/rules/src/lib.rs`'s export list and `tests/mod.rs`'s import list, merged this morning). Needs a rebase; leaving that to you rather than pushing to your branch myself.

Good to merge once rebased.

old-dov (approved):
LGTM, verified locally (54/54 tests, clippy/fmt clean) — see my comment. Approving now, will need a fresh look only if the rebase changes the diff.

Issue #216, opened 2026-09-16
watchdog: OpenRC path can orphan an unsupervised agent process after a watchdog crash


Jihair54:
Found while validating the new OpenRC support for `watchdog install` (issue #213) on the Alpine lab VM.

## Repro (Alpine, OpenRC, kernel 6.18)
1. `watchdog install` -> OpenRC starts `supervise-daemon` -> `watchdog` (layer 2) -> `agent` (layer 1), 3 distinct PIDs confirmed via `ps`.
2. `kill -9 <watchdog pid>` only (not the agent, not supervise-daemon).
3. `supervise-daemon` correctly respawns a fresh `watchdog`, which spawns a fresh `agent` -- confirmed via new PIDs and `/var/log/synthaea-watchdog.log`.
4. The old agent process (child of the killed watchdog) is still running, confirmed via `kill -0 <old agent pid>` returning alive. Two agent processes now run concurrently, both able to write to the same `alerts.ndjson`/`events.jsonl`.

## Root cause
`watchdog`'s SIGKILL bypasses its own signal handler (`supervise.rs`'s `cmd_run`), so it never gets the chance to kill its child agent before dying. The OS does not kill children when a parent dies -- the agent is simply reparented (to PID 1) and keeps running, unsupervised by anything.

On systemd this is a non-issue: `KillMode=control-group` (the default) sweeps the entire unit's cgroup -- including orphaned children -- every time the unit restarts. Confirmed this Alpine VM's OpenRC has cgroup cleanup off by default (`rc_cgroup_cleanup="NO"` in `/etc/rc.conf`, and `/sys/fs/cgroup` empty for this service), so nothing plays that role on the OpenRC path.

## Why not fixed inline
Turning on system-wide OpenRC cgroup cleanup from a per-service `watchdog install` isn't appropriate (it would affect every service on the host, not just ours). A real fix likely means process-group-based cleanup in `watchdog/src/supervise.rs` (cross-platform, not OpenRC-specific) -- a larger change than "add OpenRC support," so filing separately rather than expanding #213's scope.

## Impact
Narrow: requires the watchdog process itself (not the agent) to die in a way that skips its signal handler (SIGKILL, a host OOM-kill targeting it specifically, a hard panic). The far more common case -- the agent crashing -- is already handled cleanly by layer 1 with no orphan (confirmed on the same VM).

Jihair54:
[...]

PR #217, opened 2026-09-16
watchdog: OpenRC support for install/uninstall/status on Linux (closes #213)

### LINKED ISSUE #213: watchdog install: no OpenRC support on Linux (hard-coded to systemd)
## Problem

watchdog install on Linux (watchdog/src/service/linux.rs) is hard-coded to systemd: it writes /etc/systemd/system/synthaea-agent.service directly and shells out to the systemctl binary, with no branch for OpenRC (or any other init system).

This blocks testing "does the watchdog survive a reboot" on any OpenRC-based distro — Alpine chief among them, since Alpine never ships systemd. Same shape of gap as the existing tamper/heartbeat finding (a feature that works at the unit level but isn't wired for this environment yet) — not a regression, a missing feature.

## Evidence (Alpine lab VM, kernel 6.18.50-0-virt, main @ 38921d8)

`
$ which systemctl; echo "exit: True"
exit: 1

$ watchdog install --agent-bin /path/to/agent --alerts /root/alerts.ndjson
Error: writing /etc/systemd/system/synthaea-agent.service (root required): No such file or directory (os error 2)
`

/etc/systemd does not exist on this machine at all (no systemd installed) — the write fails with ENOENT before systemctl is even invoked.

Minor secondary nit: the error message's (root required) annotation is misleading here — that's static context text from .with_context(), not an actual diagnosis; the real cause is ENOENT, and the process is already running as root. Worth rewording if/when this code path is touched, not urgent on its own.

## Suggested next step

Add an OpenRC branch alongside the existing systemd path in watchdog/src/service/linux.rs: detect which init system is present (e.g. check for /run/openrc or /etc/systemd at install time, or a compile-time/runtime distro check), and for OpenRC generate an /etc/init.d/synthaea-agent script (supervise-daemon or a manual respawn loop, matching the systemd unit's Restart=always semantics) plus 
c-update add synthaea-agent default.

Not something to patch blindly here — real feature work, worth the team's design input on how init-system detection should work across the Linux service module before implementing.

## Related

lab/MATRIX.md documents Alpine as OpenRC-based; Roadmap_Tests_Alpine.md Phase 5 tracks this as blocking the "watchdog survives VM reboot" test case.

Jihair54:
Closes #213.

`watchdog install`/`uninstall` and `service::cmd_status` were hardcoded to systemd on Linux -- fails outright on Alpine and any other OpenRC-based distro (confirmed: no `/etc/systemd`, ENOENT on unit-file write). This adds an OpenRC path alongside the existing systemd one, detected at install time (`/run/systemd/system` vs `/sbin/openrc-run`), not assumed.

## What changed
- `watchdog/src/service/linux.rs`: `cmd_install`/`cmd_uninstall`/`cmd_status` now dispatch on detected init system. New `install_openrc`/`uninstall_openrc`/`status_openrc`, generating `/etc/init.d/synthaea-agent` (`supervisor=supervise-daemon`, `rc-update add ... default`, `rc-service ... start`).
- `watchdog/src/service/mod.rs`: `cmd_status` delegates to `linux::cmd_status` on Linux instead of hardcoding `systemctl`.

## `respawn_max` finding (not guessed -- verified on the VM)
`supervise-daemon`'s own `--help`/usage strings suggested omitting `respawn_max`/`respawn_period` gives unlimited respawn. `ps aux` after a real install showed `supervise-daemon` running with `--respawn-max 5 --respawn-period 1800` despite the generated script setting neither -- those are OpenRC's site-wide defaults from `/etc/rc.conf`'s `SUPERVISE DAEMON CONFIGURATION VARIABLES` block, applied to any service that doesn't override them. `rc.conf` documents `respawn_max=0` as unlimited, which is the real `Restart=always` equivalent -- set explicitly in the generated script so a crash loop past 5 restarts/30min (the site default) doesn't permanently kill the service.

## Testing (Alpine lab VM, kernel 6.18, real OpenRC, not mocked)
- `cargo build/test/clippy -p watchdog` clean (`-D warnings`).
- `watchdog install` -> `/etc/init.d/synthaea-agent` generated correctly, service starts, 3 distinct PIDs confirmed (`supervise-daemon` -> `watchdog` -> `agent`).
- Killing only the agent (exact PID): watchdog respawns it in place (layer 1, pre-existing code, unaffected by this change).
- Killing only the watchdog (exact PID): `supervise-daemon` respawns a fresh watchdog+agent pair (layer 2, the new code this PR adds).
- `rc-service synthaea-agent stop`: clean, no leftover processes.
- Full VM reboot (`reboot`, confirmed via `uptime`): service comes back up on its own via `rc-update add ... default` -- the reboot-survival test blocked since the start of this test phase for lack of any OpenRC path.

## Known gap, filed separately
Killing only the watchdog process leaves its old agent child as an unsupervised orphan (OS doesn't kill children when a parent dies; OpenRC has cgroup cleanup off by default on this VM, unlike systemd's `KillMode=control-group`). Fixing this properly means process-group cleanup in `watchdog/src/supervise.rs` (cross-platform, not OpenRC-specific) -- larger than "add OpenRC support," so filed as #216 instead of expanding this PR's scope.


old-dov (approved):
OpenRC support looks solid — real VM validation (kill/respawn, full reboot), honest scope discipline (orphan cleanup gap filed separately as #216). LGTM.

PR #218, opened 2026-09-16
docs/adr: 0008 — NetworkFlow feeds the correlation vector, not the behavior vector

Sollykhan:
## Summary

Records the design decision reached with @old-dov over Discord this afternoon on step (3) of my #193 review sequence (feature contract for NetworkFlow in the ML vectors):

- **NetworkFlow feeds T2 only** (correlation Isolation Forest), not T1 (BehaviorVector/BAYES). Rationale: T2 already carries the network-signal fields, T1 is positional and fragile to insertion, and one model → one verdict avoids the double-alert risk the original #193 review flagged.
- **T2 absorbs the sources**: existing \`unique_daddr_count\` / \`unique_dport_count\` counters span Connect AND NetworkFlow together, no per-source variants. Consistent with the rules-layer precedent (BEACON's \`check_beacon\` + \`check_beacon_flow\` converge on the same \`record_beacon\` state).
- **MVP scope**: extend the two existing counters only. \`flows_per_second\` / \`long_lived_flow_count\` are v2 candidates. \`bytes_sent\` features are explicitly discarded (\`nf_conntrack_acct=off\` by default). Per-port histograms are explicitly discarded (no ML value BEACON+listener-drift don't already deliver deterministically).

## Consequences captured

- Zero prod regression — no ONNX in the registry against this vector yet, so this ADR captures intent before the surface is used.
- Two Rust edits in \`correlation.rs\` documented: \`is_modeled()\` extension, \`Event::NetworkFlow\` arm in the daddr/dport loop.
- Python parity in \`correlation.py\` flagged as prerequisite before #44 generates the first NetworkFlow-inclusive baseline — otherwise the parity fixture fails on first capture.
- Documented side effect on \`is_modeled()\`: filter gates the entire vector, so \`event_count\` / \`span_s\` will also reflect NetworkFlow. \`connect_count\` in T1 is protected by an explicit \`matches!(Event::Connect(_))\` and unaffected.

Unblocks step (2) capture on the new vector, then step (4) training run, per the original #193 review sequence.

Refs #193, #44.

## Test plan

N/A — pure decision record. Format follows \`docs/adr/template.md\` (Context / Decision / Consequences), with additional Rationale / Implementation notes / Deferred sections consistent with ADRs 0002 and 0006.


old-dov:
Reviewed, "Deferred" section checked line by line against what we discussed — it's accurate:

- **Volume features excluded**: correctly ties it to `bytes_sent` being `None` by default (`nf_conntrack_acct=off`), framed as an infra decision, not a baseline problem. Matches.
- **Per-port histograms deferred, not discarded for a real reason**: correctly frames it as "no ML-side value BEACON + listener-drift don't already deliver deterministically" rather than a false "redundant computation" claim — that's the right nuance.
- **`flows_per_second`/`long_lived_flow_count` as v2, not MVP**: consistent with keeping the MVP's arity growth minimal, and both are flagged as needing a windowing/threshold decision first, which is honest about what's actually unresolved.

One factual issue outside "Deferred", in the Context/Rationale sections: the file paths for T1 are wrong. The ADR says `crates/ml/src/features/behavior.rs` and `crates/rules/src/state/bayes.rs` — I checked, neither exists. The real paths are `crates/correlator/src/behavior.rs` and `crates/correlator/src/bayes.rs`. Worth fixing before merge since this is meant to be a durable reference.

Everything else (BEACON precedent, the `is_modeled()`/`event_count`/`span_s` side-effect note, "no production regression" framing, absorb-over-coexist reasoning) matches the discussion exactly. Good to merge once the paths are corrected.

old-dov (approved):
Reviewed #218 in full — matches the decision reached on Discord. Verified the technical claims against the current tree: `is_modeled()` and `unique_daddr_count`/`unique_dport_count` exist as described in `correlation.rs`, and `check_beacon_flow` exists in the rules layer as the stated precedent. Bumped Status to accepted to match the convention of ADRs 0001-0007 before merge.

PR #221, opened 2026-09-17
docs/adr: 0009 — model record binds scenario-replay results; training.json renamed to model_record.json

Sollykhan:
## Summary

Records the design decision reached with @old-dov this morning on decision 4 of #44's scoping split (binding scenario replays to the model card).

Two coupled decisions :

1. **Extend the model record** with \`scenario_replays: list[ScenarioReplayResult]\`, content-binding each result to its source scenario YAML by \`scenario_yaml_sha256\` — same principle as the existing \`baseline_sha256\` on \`DatasetVersion\`. Passing criterion is bounded on both sides: \`min_count ≤ observed ≤ min_count + tolerance\`, per \`(technique, rule)\` exact string match, AND across all expected → \`passed\`.

2. **Rename \`training.json\` → \`model_record.json\`**. Verified pre-ADR by @old-dov: \`find . -name "training.json"\` returns nothing anywhere in the repo — the name lives only in code as \`TRAINING_RECORD_FILENAME\` in \`training_record.py:43\`, never materialised on disk. Rename now costs zero; delaying costs a real migration once real models write the file. Bump \`SCHEMA_VERSION\` 1 → 2 (v1 rejected — matches \`manifest.py\` pattern).

## Ship gate contract (explicit)

\`verify_provenance()\` at \`verify_onnx.py:159\` will enforce, in strict mode:
- \`model_record.json\` exists and \`schema_version == 2\`
- Every \`dataset_version.baseline_sha256\` matches on-disk manifest (unchanged)
- Every \`scenario_replay.scenario_yaml_sha256\` matches on-disk \`lab/scenarios/<name>.yaml\` (new)
- Every \`scenario_replay.passed\` is True, AND re-computed passing criterion agrees with the cached value (catches mutation)

Single call site, extension flows through the existing \`verify_training_record\` helper.

## Scope boundaries called out explicitly

\`card.md\` content stays out of scope. The registry README's "no card, no ship" remains a declaration of intent — same pattern as ADR-0006's policy-distribution intent doc. Mechanising a \`card.md\` existence/content check deserves its own **ADR-0010** whenever it becomes a priority. This ADR's Deferred section names that gap so a future reader does not mistake silence for mechanism.

Also deferred to v2: extended \`Environment\` (libc glibc/musl, toolchain), multi-run stability, \`render_card.py\` auto-generation of the human-readable Scenario Replay Results section.

## Consequences

- **Zero prod regression** — no ONNX in the registry is shippable today. \`cmdline-iforest-linux/0.1.0\` and \`cmdline-iforest-windows/0.1.0\` are both marked \`reference-only, NOT shippable\` and predate the training record concept entirely.
- Sets a definite contract for the upcoming trainer glue (that writes \`model_record.json\`) and the upcoming release gate (that consumes it), both of which are mentioned in \`training_record.py\`'s crate doc as future work.

Refs #44, ADR-0002, ADR-0008.

## Test plan

N/A — pure decision record. Format follows \`docs/adr/template.md\` extended with Schema / Passing criterion / Ship gate contract sections, consistent with ADR-0008.


old-dov (approved):
Reviewed #221 in full. Solid — checks out against the actual code on every factual claim I could verify:

- `verify_provenance()` really is at `verify_onnx.py:159`, and the extension (schema_version==2, scenario_yaml_sha256 match, passed re-computation) flows through the existing `verify_training_record()` call site as described, no new gate entry point.
- The v1→v2 SCHEMA_VERSION reject pattern matches `manifest.py`'s existing `if schema_version != SCHEMA_VERSION: raise` — consistent, not a new convention.
- Both placeholder cards (`cmdline-iforest-{linux,windows}/0.1.0/card.md`) really are flagged `reference-only, NOT shippable` as claimed — the "this ADR does not migrate them" consequence is accurate, not hand-waved.
- `ExpectedDetection`/exact-string `(technique, rule)` match is the right call — `Alert.technique`/`CorrelationAlert.technique` really are opaque, sometimes-combined strings (verified across `state.rs`, `stateless.rs`, `correlator/src/rules.rs` when I proposed decision 2 on #44), so decomposing would invent a taxonomy the rule engine doesn't have.
- Deferring `card.md` content verification to ADR-0010 rather than silently scoping it out is the right call — a future reader would otherwise reasonably assume this ADR mechanises "no card, no ship" in full.

No blocking issues. Two non-blocking thoughts for whenever the trainer glue actually lands (not this ADR, just flagging for later):
- The passing criterion's over-detection case (`observed_count > min_count + tolerance` → fail) will need a real tolerance default per scenario once BEACON-style rules that naturally re-fire (e.g. `record_beacon` state) start feeding `observed_detections` — worth a sentence in the trainer glue's own doc when it's written, not here.
- `Environment.kernel` as "uname -r verbatim" is Linux/macOS-shaped; worth a one-line note on what a Windows replay's `kernel` field holds (build number? `ver` output?) before the first cross-platform model record gets written, so the field doesn't end up with two different conventions by accident.

Thanks for looping in Florian's and my track explicitly (decisions 3/4 split) and for the ADR-0010 forward-reference — keeps #44 from growing an unbounded tail.

PR #229, opened 2026-09-18
lab/scenarios: convert Linux scripts to YAML sidecars (issue #44)

old-dov:
## Summary

Applies decisions 1 & 2 posted on #44 (2026-09-17) to the six Linux scenarios:
a YAML sidecar per scenario (`lab/scenarios/<name>.yaml`) and an
`ExpectedDetection{technique, rule, min_count, tolerance}` schema matching
`Alert.technique`/`CorrelationAlert.technique` by exact string.

- `beacon.yaml`, `dropper-chain.yaml` — as sketched in the #44 thread.
- `argv.yaml`, `dns-exfil.yaml`, `lineage.yaml`, `respawn-beacon.yaml` — new,
  the "other 4" from my 2026-09-17 comment.

`technique`/`rule` values were checked against the actual constants, not just
the scripts' header comments:

- `beacon.sh` → `T1071/T1041`, `check_beacon` (`crates/rules/src/state.rs`).
- `dropper-chain.sh` → `T1105/T1059/T1071`, `rule_spawn_connect_filewrite`
  (`crates/correlator/src/rules.rs`).
- `respawn-beacon.sh` → **two** expected detections, not one: `check_beacon`
  (`T1071/T1041`, client-side connections cross the threshold regardless of
  listener pid churn — keyed on `(comm, daddr, dport)`) **and**
  `rule_respawn_connect` (`T1059/T1071`, keyed on `(ppid, comm)`). The
  script's original header documented this as an open hypothesis from
  2026-08-25; `rule_respawn_connect`'s own doc comment says it was since
  confirmed and fixed during the edr-new migration, so both alerts are now
  expected to fire reliably.
- `lineage.sh` / `argv.sh` → `T1059` / `T1059.004`
  (`check_web_server_spawns_shell` / `check_base64_decode`), `min_count: 3`
  since both scripts hardcode `ITERATIONS=3` with one alert per iteration and
  neither rule does windowed dedup.
- `dns-exfil.sh` → `T1048.003/T1071.004`, `rule_dns_exfil`. Tagged
  `platform: windows`, not linux: there's no Linux DNS sensor yet (script
  comment), so the scenario is a no-op unless run with the agent on a Windows
  lab host — the driving script is POSIX but the *observing* platform is
  Windows.

## `.gitattributes`

Pinned `lab/scenarios/*.yaml` to LF. ADR-0009 (merged after the decisions
were posted, #221/#222) hashes these files via
`ScenarioReplayResult.scenario_yaml_sha256` to bind a model record to the
exact scenario it was replayed against — same CRLF hash-drift risk on a
Windows checkout that #196/#198 already fixed for `ml/datasets/**`.

## Scope

Format + conversion only, no replay-engine code — that consumes decision 2
and belongs with the Python entry point on @Sollykhan's track, per the
original split. Decisions 3 (storage backend, Florian) and 4 (model-card ↔
dataset-version binding) are untouched and still open on #44.

@Sollykhan — flagging you specifically since ADR-0009 (yours) is the first
consumer of this exact schema shape (`ExpectedDetection`/`ObservedDetection`
field names match what you specced there) and the `.gitattributes` pin
protects the `scenario_yaml_sha256` binding it introduces. Worth a look
before the replay engine starts reading these.

Refs #44.


old-dov:
Ran all 6 scenarios against a live \`agent run\` (WSL2, kernel 6.6.114-microsoft-standard-WSL2) instead of leaving these as source-reading guesses. Pushed corrections in 9b39bde.

**Confirmed as declared:** \`beacon\`, \`dropper-chain\`, \`dns-exfil\` all matched their \`expected_detections\` exactly.

**Corrected — \`respawn-beacon\`:** dropped \`rule_respawn_connect\` from the assertions. Only \`sys_enter_connect\` is probed, never \`accept()\`, so the listener's own \`(ppid, comm)\` bucket has no \`Connect\` event of its own to pair with its \`Exec\` respawns — the rule needs both in the *same* bucket. The 4 firings I saw in this run line up better with a WSL2 defect (every process reported \`ppid=0\`, merging the listener's respawns and the client's unrelated connections into one bucket) than with the rule actually working for this scenario shape. @Sollykhan — this rule's doc comment says the respawn hypothesis was "confirmed," but that traces to a unit-test re-read (same-pid-reused test), not a live run of \`respawn-beacon.sh\`. Worth a real-kernel check (Hyper-V lab) on whether it fires at all here before either scenario or rule changes. Full reasoning in the yaml's \`notes\`.

**Could not validate — \`lineage\`, \`argv\`:** every event in this WSL2 run showed \`ppid=0\` and empty \`cmdline\`/\`argv\`, which is exactly what these two scenarios test for — so I can't tell a real regression from an environment defect here. WSL2 isn't a \`lab/MATRIX.md\` row. Left \`min_count\` as reasoned from source, flagged both as unconfirmed pending a real-kernel run.

old-dov:
Correction to my own comment above: \`lineage\`/\`argv\` aren't actually unconfirmed — I'd forgotten these were already validated on real hardware on 2026-09-09/10 via the Hyper-V lab (Ubuntu 22.04/5.15.0-91, Debian 12/6.1.0-17, PR #155): 3/3 alerts on both kernels for each, matching \`min_count: 3\` exactly. Fixed the notes in ec0c07e to say that instead of "unconfirmed" — the WSL2 run today just hit an unrelated environment defect (WSL2 isn't a MATRIX.md row), not evidence against the scenarios.

Sollykhan (approved):
Approuvé sur la partie schema — les noms de champs `ExpectedDetection`/`ObservedDetection` (technique, rule, min_count, tolerance) matchent exactement ce que j'ai spéc dans ADR-0009. Le `.gitattributes` LF est bien vu, il protège directement `scenario_yaml_sha256` du drift CRLF sur checkout Windows — c'est le risque que j'avais nommé "documenté mais pas mécanisé" dans l'ADR, tu viens de le mécaniser.

Sur ton run WSL2 (ppid=0 + argv/cmdline vide sur tous les exec events) : je suis tombé sur exactement le même comportement hier après-midi en capturant une baseline (7427 events tous avec argv=[]). Après lecture du probe eBPF (`crates/sensors/linux/ebpf/src/main.rs`) et du userspace loader (`normalize.rs`), c'est bien issue #152 fermée (Option 4 : userspace lit `/proc/<pid>/cmdline` au drain, racy pour processus courts). WSL2 amplifie parce que curl/sleep meurent avant le drain userspace. Pas un WSL2 defect au sens strict, plutôt le compromis architectural qui se manifeste massivement sur kernel WSL2 + workloads courts. J'allais te ping ce matin pour comparer nos observations — tu m'as devancé.

Sur `rule_respawn_connect` : point pris, tu as raison, le "confirmed" du doc comment traçait sur un re-read de test unitaire pas un run réel. Je le mets sur ma liste vérif Hyper-V — je récupère un MacBook la semaine prochaine et enchaîne avec une VM Hyper-V, je te confirme là.

PR #230, opened 2026-09-18
fix(watchdog): kill the agent when the watchdog is SIGKILLed (closes #216)

### LINKED ISSUE #216: watchdog: OpenRC path can orphan an unsupervised agent process after a watchdog crash

Jihair54:
## Summary
- A watchdog killed by SIGKILL skips its own signal handler, so the agent it spawned is never explicitly stopped — it just gets reparented to PID 1 and keeps running, unsupervised, once the service manager respawns a fresh watchdog+agent pair. Confirmed on Alpine/OpenRC, which (unlike systemd's default `KillMode=control-group`) does not sweep orphaned children on restart.
- Fix: register `PR_SET_PDEATHSIG(SIGKILL)` on the agent child right after `fork` (before `exec`, via `pre_exec`), so the kernel kills it the instant its parent dies, regardless of how the parent died. Guards the fork/prctl race by checking `getppid()` hasn't already changed before proceeding to exec. Linux-only (`#[cfg(target_os = "linux")]`) — systemd already covers this via cgroups, and this is the OS-level primitive that closes the same gap on OpenRC.
- Regression test (`supervise::pdeathsig_tests::agent_dies_when_watchdog_is_sigkilled`, Linux-only): re-execs the test binary as a stand-in "watchdog" that spawns a grandchild through the same `die_with_parent` path, then SIGKILLs the stand-in and asserts the grandchild disappears on its own.

## Test plan
- [x] `cargo test -p watchdog --release` — 16 passed, including the new regression test.
- [x] `cargo clippy -p watchdog --all-targets -- -D warnings` — clean.
- [x] Real end-to-end repro on the Alpine lab VM (OpenRC): installed the `synthaea-agent` service, confirmed 3 PIDs (`supervise-daemon` → `watchdog` → `agent`), `kill -9`'d the watchdog PID only — the agent died immediately (previously it survived as an orphan), and `supervise-daemon` respawned a clean new watchdog+agent pair with no leftover processes.

Closes #216.


old-dov (approved):
LGTM. `PR_SET_PDEATHSIG` set in `pre_exec` after fork/before exec is the right primitive for this — no thread/heap state in scope at that point, matches the SAFETY comment. The `getppid() != watchdog_pid` check correctly closes the fork/prctl race (already-reparented child bails instead of exec-ing unsupervised).

Regression test is a good design: re-exec-as-fake-watchdog + real SIGKILL reproduces the exact failure mode (bypasses signal handlers entirely, unlike a graceful stop), not just a unit-level assertion on the syscall. Real Alpine/OpenRC end-to-end repro in the test plan closes the loop (matches why systemd's `KillMode=control-group` doesn't need this but OpenRC does).

`libc` scoped to `cfg(target_os = "linux")` only, small diff, nothing to flag. Approving.

PR #232, opened 2026-09-18
lab/scenarios: correct rule_respawn_connect after real-hardware validation

old-dov:
## Summary

#229 removed `rule_respawn_connect` from `beacon.yaml`/`respawn-beacon.yaml`'s
`expected_detections`, based on a WSL2 test where the alert fired but looked like
an artifact of WSL2's known ppid=0 defect. Ran both scenarios today on real
hardware (Hyper-V lab, Arch Linux `7.2.4-arch1-2`) to settle it properly rather
than leave the question open.

## What's actually going on

The alert fires reliably on real hardware too — 2/2 for each script — but not
for the reason either script's own header comment (or my previous note)
assumed. Traced the exact pids on the VM:

- The **listener's** own respawns (`setsid bash -c 'while true; do nc -l ...;
  done'`) share one `(ppid, comm)` group and confirmed **never** produce a
  `Connect` event — `agent status` only lists `sys_enter_connect`, no `accept()`
  probe. That part of the earlier analysis holds: the listener side alone
  genuinely cannot satisfy `rule_respawn_connect`.
- What actually trips it is each script's **client** loop:
  `for i in ...; do echo ... | nc -w1 ...; done` spawns N `nc` invocations that
  all share the *script's own pid* as parent — a second, independent
  `(ppid, comm)` group with both repeated `Exec` (>= `RESPAWN_THRESHOLD`) and
  `Connect` in the same bucket. That's a real, general pattern, not something
  specific to WSL2.

## Change

Re-added the detection to both `beacon.yaml` and `respawn-beacon.yaml`
(`min_count: 1`, `tolerance: 1` — the exact alert count is timing-sensitive,
not worth asserting a fixed number). Notes on both explain the real mechanism
and flag that it doesn't match what either script's comments describe —
worth a follow-up to fix the framing or redesign the scenario (e.g. `setsid`
each client connection so it doesn't share the script's pid) if testing the
listener-respawn case specifically still matters. Not done here.

## Also validated on this VM (separate from this diff)

All three `Done when` items on issue #80 (container attribution) now pass end
to end: container id/image/name attribution on exec events, beacon detection
from inside a container with attribution visible on the resulting `Connect`
events, and the T1611 proc-root-escape rule. Posted as a comment on #80
rather than a code change, since #80's code already merged weeks ago — this
was pure validation.

Refs #44, #80, #92.


SherlockOmss (approved):
## Review Summary

Solid investigation with real-hardware validation that corrects the premature removal from PR #229. The root-cause analysis is correct and thoroughly documented.

**✅ Approval reasons:**
- Rigorous validation on real hardware (Arch Linux Hyper-V), not just WSL2
- Accurate identification of the real trigger (client loop sharing script's ppid, not listener respawns)
- Honest correction of the WSL2 misdiagnosis
- Appropriate tolerance for timing-sensitive alert counts
- Comprehensive notes explaining what actually fires vs what the script headers claim

**📋 Non-blocking suggestions:**

**1. Follow-up issue needed**

The notes correctly flag that the script headers describe testing "listener respawning" but the rule actually fires due to the client loop pattern. This deserves a tracked follow-up:

- **Option A:** Update script headers to describe what's actually tested (client loop with shared ppid)
- **Option B:** Redesign scenarios to test the listener-respawn case specifically (e.g., `setsid` each client connection)

Recommend creating an issue and referencing it in the YAML notes.

**2. Clarify "2 alert lines" observation**

`respawn-beacon.yaml` mentions:
```yaml
tolerance: 1   # observed 2 alert lines on real hardware (see notes)
```

The notes don't explain why 2 lines were observed. Is this:
- The rule firing twice (once per ppid/comm group)?
- A timing edge case?

Adding a sentence in the notes would help future readers understand the tolerance choice.

**3. Accept() tracing limitation**

The analysis reveals that `rule_respawn_connect` cannot detect the "listener that respawns after each connection" pattern because:
- Only `sys_enter_connect` is probed, not `accept()`
- Listeners never call `connect()`, only `accept()`

If this detection gap matters for real-world implant behavior, consider documenting it (ADR? issue for future sensor work?). Not blocking for this PR.

**4. Minor wording improvements:**

**beacon.yaml line 23-24:**
```diff
- CONNECTIONS (4) `nc` invocations that all share the *script's own* pid as parent —
- that alone is >= RESPAWN_THRESHOLD (3) with each spawn also connecting
+ CONNECTIONS (4) `nc` invocations >= RESPAWN_THRESHOLD (3), all sharing the 
+ *script's own* pid as parent — with each spawn also connecting
```

Makes the threshold comparison more explicit.

**Both files: add line reference to correlator rule:**
```yaml
rule_respawn_connect groups by (ppid, comm) — see crates/correlator/src/rules.rs:135-149
```

---

**Validated cross-issue:** Good to see issue #80 (container attribution) validation on the same VM mentioned in the PR body. Shows thorough lab usage.

**Verdict:** LGTM. The technical correction is sound, documentation is excellent. Suggestions above are for polish, not blockers.

PR #234, opened 2026-09-18
fix(sensor-linux): resolve container attribution from a kernel-captured cgroup id (closes #204)

### LINKED ISSUE #204: sensor-linux: container attribution loses the exit race for single-shot processes, silently missing container-conditioned detections
## Summary

Container attribution (`container_context` / `read_container_id`,
`crates/sensors/linux/userspace/src/sensor.rs`) resolves a pid's container
id lazily, by reading `/proc/<pid>/cgroup` when the event is drained from
the ring buffer — not at the moment the kernel probe captures the event.
For a process whose entire lifetime is one `open()` then exit (e.g. a bare
`cat <path>`, as opposed to a shell that keeps running afterward), this
reliably loses the race: by the time the drain loop gets to it, the pid's
`/proc` entry is already gone, on the very first attempt, every time.

**Practical impact:** any container-conditioned detection rule silently
misses a one-shot command. Found while validating #80's `check_proc_root_escape`
(T1611, Escape to Host) end-to-end against a real Docker daemon: a
`docker exec <container> cat /proc/1/root/etc/hostname` (BusyBox/Alpine or
GNU coreutils, same result) genuinely succeeds — the host's own hostname
gets printed — but produces zero container attribution and the rule never
fires, even though the exact same access via a process that stays alive a
moment longer (`sh -c 'exec 3</proc/1/root/etc/hostname; sleep 3'`) is
attributed correctly and does fire the alert.

## Root cause

`ContainerIdCache::resolve_with` (`sensor.rs`) already retries a `None`
result up to `NONE_RETRY_LIMIT` (5) times before trusting it as final —
that mechanism was built for a *different* race (`runc`'s init process
doing several `file_open`s before it moves into the container's cgroup
and `execve`s into the target command, found while building #80). It does
not help here: every retry attempt within the same drained-event burst
sees the identical `ENOENT` on `/proc/<pid>/cgroup`, because the process
is already gone from the very first attempt, not just slow to settle.
Confirmed live with a temporary debug log — for the affected pid, attempt
0 through 32 (the full burst of `file_open`s for that pid) all returned
`is_proc_exit_race(&e) => true` (`ENOENT`), never once succeeding.

The underlying issue is architectural: attribution is resolved
drain-time, against `/proc`, which requires the process to still exist
at that point. A process whose whole lifetime is under the userspace
drain loop's latency loses every time, deterministically — this isn't a
flaky race that occasionally fails, it's one that occasionally *doesn't*.

## Repro

```bash
# against a running agent + a --pid=host container (see #80's PR for setup)
sudo docker exec <container> cat /proc/1/root/etc/hostname
# succeeds (prints the host's hostname) — but:
grep '/proc/1/root' events.jsonl | tail -1
# → file_open event for this pid has no "container" field at all
grep T1611 alerts.ndjson
# → nothing

# same access, but the pid stays alive past the open:
sudo docker exec <container> sh -c 'exec 3</proc/1/root/etc/hostname; sleep 3'
# → file_open AND exec events for this pid both carry the correct
#   container id/image/name, and check_proc_root_escape fires T1611
```

## Evidence

Validated on the Arch Hyper-V VM from #124, against a real `dockerd`
(`docker.io 29.1.3`), with `debian:12-slim` (ruling out any BusyBox/Alpine
relative-`openat()` quirk — see #80's PR discussion — this is unrelated to
that; the path here is a single absolute `openat()` and still isn't
attributed in time).

## Suggested fix directions (not decided here)

- Capture the container id kernel-side, at the point the eBPF probe fires
  (e.g. `bpf_get_current_cgroup_id()`), and carry it in the wire event
  itself — removes the drain-time race entirely, since the read would
  happen while the process is still executing the syscall. Bigger change:
  touches the eBPF program and the wire format, not just userspace.
- Accept the gap and document it as a known limitation for
  container-conditioned rules (already done on `read_container_id` and
  `check_proc_root_escape`'s doc comments as of #80) — cheapest, but means
  a one-shot escape/exfil command via a fast, short-lived process is
  invisible to every rule that depends on `meta.container`.

Filed while validating #80's `check_proc_root_escape` — not a bug in that
rule's logic (6/6 unit tests pass; the gap is entirely in what attribution
data reaches it), and not specific to T1611: any future rule gated on
`meta.container.is_some()` inherits the same blind spot.

Jihair54:
## Summary

Container attribution lost the exit race for single-shot processes (e.g. a bare `docker exec <c> cat <path>`): the sensor read `/proc/<pid>/cgroup` at drain time, after the pid could already be gone, silently dropping container context on any container-conditioned detection (e.g. T1611 `check_proc_root_escape`).

- Captures the process's cgroup id kernel-side, at syscall time, via `bpf_get_current_cgroup_id()` in all three eBPF probes (exec, file_open, connect) — added to `EventMeta` as `cgroup_id: u64` (`sensor-linux-wire` `WIRE_VERSION` bumped 3→4).
- Userspace resolves the container id from that cgroup id by walking `/sys/fs/cgroup` for the directory whose inode matches (`container_id_from_cgroupfs`, same technique as Falco/Tracee) — the cgroup directory persists for the container's entire lifetime, independent of any single process inside it, so the race is eliminated rather than mitigated.
- Replaced the old pid-keyed `ContainerIdCache` (which cached `/proc/<pid>/cgroup` reads and needed retry/NONE_RETRY_LIMIT bookkeeping to paper over the race) with a `CgroupIdCache` keyed on the stable cgroup id.

## Test plan

- [x] `cargo build --release -p sensor-linux -p sensor-linux-wire -p rules` — clean, confirms `bpf_get_current_cgroup_id()` against the pinned aya-ebpf version.
- [x] `cargo test --release -p sensor-linux -p rules` — 46/46 + 74/74 passed, including new `cgroup_id_cache_*` / `cgroupfs_walk_*` unit tests.
- [x] `cargo clippy -p sensor-linux -p rules --all-targets -- -D warnings` — clean (pre-existing unrelated `bpf_target_arch` build-script warnings only).
- [x] Real end-to-end repro on the Alpine lab VM against a live Docker daemon (snapshotted VM before install): started a container (`--pid=host`), ran `docker exec probetest cat /proc/1/root/etc/hostname` twice — both single-shot `cat` invocations exit almost immediately, the exact race this issue describes. Both correctly produced a T1611 alert with the right `container=<id>` attribution (2/2, no missed detections), confirmed via `agent run --alerts alerts.ndjson --events events.jsonl`.

Closes #204.


SherlockOmss (approved):
## Review Summary

Excellent fix for issue #204's container attribution race. The kernel-side capture approach is the correct solution and eliminates the race entirely rather than just mitigating it.

**✅ Approval reasons:**
- Correct root-cause fix: captures cgroup_id at syscall time (kernel-side) before the process can exit
- Proper technique: `bpf_get_current_cgroup_id()` + cgroupfs inode walk matches Falco/Tracee patterns
- Simplified cache: removed retry logic (no longer needed since cgroup_id is already "settled" at capture time)
- Comprehensive testing: real end-to-end repro on Alpine VM with Docker (2/2 T1611 alerts with correct attribution)
- Clean documentation: updated doc comments to explain the race closure

---

## Technical Deep Dive

### ✅ Core Architecture Change

**Before (issue #204):**
1. eBPF captures event with metadata (pid, uid, comm, etc.)
2. Userspace drains ring buffer
3. Userspace reads `/proc/<pid>/cgroup` **at drain time**
4. **Race:** If process already exited, `/proc/<pid>/cgroup` doesn't exist → attribution silently dropped

**After (this PR):**
1. eBPF captures event **AND** `cgroup_id = bpf_get_current_cgroup_id()` **at syscall time**
2. Userspace drains ring buffer
3. Userspace walks `/sys/fs/cgroup` to find directory whose inode == `cgroup_id`
4. **No race:** Cgroup directory persists for container's entire lifetime, independent of any single process

### ✅ WIRE_VERSION Bump (3→4)

**crates/sensors/linux/wire/src/lib.rs:**
```rust
pub struct EventMeta {
    pub pid: u32,
    pub ppid: u32,
    pub uid: u32,
    pub gid: u32,
    pub timestamp_ns: u64,
    pub comm: [u8; TASK_COMM_LEN],
+   pub cgroup_id: u64,  // NEW: bpf_get_current_cgroup_id() captured at probe time
}
```

Tripwire correctly updated in `normalize.rs`:
```rust
-const _: () = assert!(wire::WIRE_VERSION == 3);
+const _: () = assert!(wire::WIRE_VERSION == 4);
```

### ✅ eBPF Probes Updated

All three probes now capture cgroup_id:
- `try_sched_process_exec` (line 185)
- `emit_file_open_event` (line 349)
- `try_sys_enter_connect` (line 471)

```rust
(*e).meta.cgroup_id = aya_ebpf::helpers::bpf_get_current_cgroup_id();
```

**Note:** Helper returns 0 on failure (e.g., kernel too old) - handled gracefully by userspace (skips cgroupfs walk if cgroup_id == 0).

### ✅ CgroupIdCache Simplification

**Old ContainerIdCache (pid-keyed):**
- Needed retry logic (`NONE_RETRY_LIMIT = 5`) for runc init window race
- Cached `None` results required `none_attempts` counter
- Complex logic to handle "not settled yet" vs "really None"

**New CgroupIdCache (cgroup-id-keyed):**
- **No retry logic needed** - cgroup_id captured at syscall time is already "settled"
- Simple HashMap<u64, Option<String>>
- Cleaner code: removed ~70 lines of retry bookkeeping

```rust
fn resolve(&mut self, cgroup_id: u64) -> Option<String> {
    if cgroup_id == 0 {  // Helper failed, skip expensive walk
        return None;
    }
    if let Some(cached) = self.entries.get(&cgroup_id) {
        return cached.clone();
    }
    // First sighting: walk cgroupfs once, cache result forever
    let id = container_id_from_cgroupfs(Path::new(CGROUPFS_ROOT), cgroup_id);
    self.entries.insert(cgroup_id, id.clone());
    // ... eviction logic ...
}
```

### ✅ cgroupfs Walk Implementation

```rust
fn container_id_from_cgroupfs(cgroupfs_root: &Path, cgroup_id: u64) -> Option<String> {
    fn walk(dir: &Path, cgroup_id: u64, depth: u8) -> Option<String> {
        if depth > MAX_WALK_DEPTH { return None; }  // Bound recursion
        for entry in std::fs::read_dir(dir).ok()?.flatten() {
            let metadata = entry.metadata().ok()?;
            if !metadata.is_dir() { continue; }
            
            // Match by inode (cgroup_id IS the directory's inode)
            if metadata.ino() == cgroup_id {
                return extract_container_id(&entry.path().to_string_lossy());
            }
            
            // Recurse into subdirectories
            if let Some(id) = walk(&entry.path(), cgroup_id, depth + 1) {
                return Some(id);
            }
        }
        None
    }
    walk(cgroupfs_root, cgroup_id, 0)
}
```

**Technique:** Standard approach used by Falco/Tracee - `bpf_get_current_cgroup_id()` returns the directory's inode under `/sys/fs/cgroup`.

**Performance:** Bounded by `MAX_WALK_DEPTH = 12` + cached after first lookup. Cgroup trees are shallow in practice (handful of slice/scope levels).

### ✅ Documentation Updates

**rules/src/stateless.rs - check_proc_root_escape:**
```rust
-/// Coverage gap, confirmed against a real Docker daemon: this needs `event.meta
-/// .container` to be populated, and the sensor's attribution loses a race for a
-/// process whose entire lifetime is one `open()` then exit (a bare `cat <path>`)
+/// Needs `event.meta.container` to be populated. Originally had a coverage gap here,
+/// confirmed against a real Docker daemon: the sensor's attribution lost a race for a
+/// process whose entire lifetime was one `open()` then exit (a bare `cat <path>`),
+/// because attribution was read from `/proc/<pid>/cgroup` at drain time, after the pid
+/// could already be gone. Closed by issue #204 — attribution is now keyed off a cgroup
+/// id captured kernel-side at syscall time
```

**Excellent:** Preserved the historical context while documenting the fix.

---

## Testing Validation

**Unit tests (46 + 74 passed):**
- ✅ `cgroupfs_walk_finds_a_matching_docker_scope_by_inode`
- ✅ `cgroupfs_walk_returns_none_if_no_match`
- ✅ `cgroupfs_walk_prefers_first_match_in_depth_first_order`
- ✅ `cgroup_id_cache_*` tests

**End-to-end (Alpine VM, real Docker):**
- ✅ Started container with `--pid=host`
- ✅ Ran `docker exec probetest cat /proc/1/root/etc/hostname` twice
- ✅ Both single-shot cat invocations exited immediately (exact race condition)
- ✅ **2/2 produced T1611 alerts with correct container attribution**
- ✅ No missed detections

**Verdict:** Real-world validation confirms the race is fixed.

---

### 1. cgroup v1 fallback

Doc comment mentions:
> A host running cgroup v1 only would not find a match here — not addressed, the same "known limitation, not this issue's scope" posture

**Acceptable:** Labs target modern kernels (Alpine/Debian/Arch with cgroup v2). Documenting the limitation is sufficient.

### 2. Performance consideration

Walking `/sys/fs/cgroup` on every first-sighting is more expensive than the old `/proc/<pid>/cgroup` read, but:
- ✅ Happens **once per cgroup** (not once per process) - far fewer cache misses in practice
- ✅ Bounded by `MAX_WALK_DEPTH = 12`
- ✅ Old approach had `NONE_RETRY_LIMIT = 5` overhead for every bare-metal process
- ✅ Eliminates the issue #199 self-sustaining loop (file_open events triggering more file_open events)

**Net:** Likely wash or improvement in real workloads.

### 3. Helper failure (cgroup_id == 0)

```rust
if cgroup_id == 0 {
    return None;
}
```

**Good:** Skips expensive cgroupfs walk when helper unavailable (old kernels). Silently degrades to no container attribution, same as before.

---

## Verdict

**LGTM - merge immediately.** This is a clean, correct fix for a confirmed real-world race condition with comprehensive testing and proper documentation.

**Impact:**
- ✅ Closes issue #204 (container attribution race)
- ✅ Enables T1611 detection for ephemeral container processes
- ✅ Simplifies cache logic (removes retry complexity)
- ✅ No regressions (backward compatible, helper failure handled gracefully)

**Recommendation:** Document this as a case study in the project's architecture docs - excellent example of "capture at the source" solving races that lazy resolution can't fix.

PR #235, opened 2026-09-18
fix(correlator): exclude wget/chronyd from BAYES false positives (closes #212)

### LINKED ISSUE #212: BAYES correlator: high-confidence false positives on ordinary activity (wget, chronyd)
## Problem

crates/correlator/src/bayes.rs's Naive Bayes filter, wired into gent run on Linux, produces high-confidence false positives on completely ordinary, unprompted system activity. Found while running Phase 2 (detection pipeline) and Phase 6 (sinks) of the Alpine lab test roadmap.

## Evidence

**Case 1 — wget to a public IP (deliberate benign-activity test):**
`
wget -T 3 -O /dev/null http://1.1.1.1/
`
produced:
`
{"technique":"BAYES","message":"pid=3736 comm=wget: high Bayesian score (log_odds=5.33, P=100%)"}
`

**Case 2 — chronyd (Alpine's stock NTP daemon, never invoked by the tester):**
`
[ALERT] BAYES — pid=2229 comm=chronyd: high Bayesian score (log_odds=2.48, P=92%)
[ALERT] BAYES — pid=2229 comm=chronyd: high Bayesian score (log_odds=2.90, P=95%)
[ALERT] BAYES — pid=2229 comm=chronyd: high Bayesian score (log_odds=2.04, P=89%)
`
This one is the more concerning of the two: chronyd is a default-installed system service, not something the tester ran. Any default Alpine install would surface these alerts with zero user action.

## Suspected root cause

log_likelihood_ratio in ayes.rs is "calibrated empirically on the NjRAT capture (
_benign=97, n_malicious=23) on 2026-08-28" — a small, single-family calibration set. Several of its per-feature weights combine to trivially match ordinary network-client behavior:

- 	ime_exec_to_connect_ms < 2000 → +1.90 (any tool that connects shortly after starting)
- dest_is_external → +1.23 (any connection to a public IP/NTP pool)
- connect_count in certain buckets → up to +2.22 (repeated/periodic connections — legitimate for both a DNS-resolving multi-A-record client like wget to 1.1.1.1, and for a periodic service like chronyd's resync cycle)

None of these features individually distinguish "malware beaconing" from "any CLI network tool" or "any periodic system service" — the benign calibration set apparently didn't include representative examples of either pattern.

## Suggested next step

Not something to patch blindly (recalibrating LLR weights is a data/product decision, not a code fix) — flagging for the team to decide: expand the benign calibration set with ordinary CLI tools (wget/curl/ssh) and common system daemons (chronyd, sshd, dhcpcd, etc.), and/or add explicit exclusions analogous to BEACON's STANDARD_PORTS/BROWSERS name lists in crates/rules/src/state.rs.

## Environment

Alpine Linux lab VM, kernel 6.18.50-0-virt, main @ 38921d8 (2026-09-16), tested via gent run end-to-end (not a unit test finding).

Jihair54:
## Summary

The Naive Bayes correlator (`crates/correlator/src/bayes.rs`) fired high-confidence alerts on ordinary, unprompted activity, found while running the Alpine lab test roadmap:

- A bare `wget -T 3 -O /dev/null http://1.1.1.1/` alone produced `log_odds=5.33` (P=100%).
- `chronyd`, Alpine's stock NTP daemon (never invoked by the tester), crossed the threshold from its periodic resync connects alone (`log_odds` up to 2.90, P=95%).

Root cause: `time_exec_to_connect_ms` (quick connect after spawn) and `dest_is_external` don't distinguish "malware beaconing" from "any CLI network tool" or "any periodic system service" — the empirical calibration set (`n_benign=97`, single-family NjRAT capture) didn't include representative examples of either pattern.

Recalibrating the LLR weights is a data/product decision, not something to patch blindly here. Instead, following the same approach already used for BEACON's `STANDARD_PORTS`/`BROWSERS` exclusions:

- Added `BAYES_NAME_EXCLUSIONS` (`wget`, `chronyd` — the two cases with lab evidence) in `crates/correlator/src/engine.rs`, narrower in scope than the existing `IGNORED` list: only the Bayesian alert is suppressed, the co-occurrence rules (`rule_spawn_connect`, `rule_dns_exfil`, etc.) still run normally against these comms.
- Guarded by the same masquerade check `IGNORED` already uses (`policy::name_exclusion_applies` on the exec's image path) — a payload renamed to `wget` or `chronyd` from an untrusted path (e.g. `/tmp/wget`) keeps full Bayesian scoring.

## Test plan

- [x] `cargo test -p correlator` — 44/44 passed, including 3 new regression tests:
  - `wget_from_a_trusted_path_never_alerts_bayes`
  - `chronyd_without_an_observed_exec_never_alerts_bayes` (chronyd predates the agent — no `ExecEvent` is ever observed for it, confirmed by reading `prime_proc_lineage` in `sensor-linux`)
  - `masqueraded_wget_from_an_untrusted_path_still_alerts_bayes` (anti-bypass regression, same pattern as the existing `masqueraded_ignored_name_is_still_correlated`)
- [x] `cargo clippy -p correlator --all-targets -- -D warnings` — clean.
- This crate has no platform-specific code (correlator depends only on the base tier, never a sensor), so the fix is pure, portable Rust logic — validated on the Alpine VM, no eBPF/toolchain involved.

Closes #212.


old-dov (approved):
Clean fix — matches the existing IGNORED/masquerade-detection pattern exactly rather than inventing a parallel mechanism, and correctly narrower in scope (only the BAYES alert is suppressed, `self.evaluate()`'s co-occurrence rules still run for wget/chronyd, matching the PR description).

Traced the two code paths: `is_bayes_excluded` reuses the same backslash-strip + case-insensitive match as `is_ignored`, and the masquerade gate is wired the same way (`self.masquerading` insert on `(is_ignored || is_bayes_excluded) && (path not trusted || parent not expected)`, checked via `peek` before suppressing). The chronyd test correctly exploits that `self.masquerading` is only ever populated from an Exec event — chronyd predates the agent so that never fires, and the gate still holds since `is_bayes_excluded` alone is sufficient when nothing marked it as masquerading.

Anti-bypass regression (`masqueraded_wget_from_an_untrusted_path_still_alerts_bayes") closes the obvious hole. LGTM.

Issue #136, opened 2026-09-04
Windows minifilter file telemetry: delete/rename, ADS, timestomping, raw volume


florianamette:
Split out of the driver umbrella (#39). The file-modification events ETW does **not** provide (audit F-6): deletes, renames, alternate data streams, timestomping via `SetInformation`, and raw-volume access.

**Directly needed by:** ransomware detection (#82 — mass rename/delete bursts are the earliest encryption signal) and timestomping rules. Today Windows is blind to file deletes/renames, so these detections cannot reach Linux-level fidelity.

Runs on the kernel-driver framework (#39); split out so it can be scoped and validated as a telemetry source independent of the driver's other capabilities.

florianamette:
Dependency note: blocked by #39 — a minifilter is a component of the kernel driver; there is no user-mode path to this telemetry.

Issue #39, opened 2026-09-03
Windows kernel driver (minifilter, ELAM/PPL) — long term


florianamette:
**The signed kernel-driver vehicle** that hosts the minifilter and kernel callbacks — the infrastructure its telemetry capabilities depend on. Long-term: gated on the code-signing / attestation-signing (MVI) path.

**This issue = the driver framework:**
- Code signing / attestation-signing; ELAM + PPL launch (tamper-resistant, protected-process).
- The minifilter framework and kernel-callback registration the sources plug into.
- Process/image/registry kernel callbacks.

**Telemetry capabilities split into their own issues (Collection · Windows), each gated on this framework:**
- #136 minifilter file telemetry (delete/rename, ADS, timestomping, raw volume)
- #137 handle-access + injection (ObRegisterCallbacks / LSASS, TI-ETW)
- #138 WFP network telemetry + inline block

Kept here (M6) as the resilience vehicle — a PPL/ELAM driver is also what makes the agent itself kernel-tamper-resistant. Unblocks #82 ransomware fidelity, #85 Windows memory scanning, and handle-access credential-theft detection once the signing path opens.

florianamette:
Scope addition from the telemetry source inventory (docs/sensors/sources.md): the driver must also register ObRegisterCallbacks for HANDLE-ACCESS telemetry — cross-process handle opens to sensitive targets (LSASS first) are the canonical credential-dumping signal, and the old README's capability matrix already promised handle-access telemetry. Add to the driver's capability list alongside minifilter + TI-ETW.

florianamette:
Classic-EDR parity additions to the minifilter scope (docs/sensors/sources.md): file creation-time changes (timestomping, Sysmon EID 2 equivalent), Alternate Data Stream writes/executes, and raw volume/PhysicalDrive access (bootkit/wiper precursor — near-zero-FP). Linux counterparts (utimensat/inode_setattr timestomping) noted on the sensor-linux-lsm scope (#91).

PR #236, opened 2026-09-18
docs/adr: 0012 - Windows driver framework language choice

SoFrexs:
## Summary

Scoping ADR for #39 (Windows kernel driver framework), which #136/#137/#138 all depend on. Separates two things #39 conflates: ELAM/attestation signing (genuinely gated on MVI, an org-level process) vs. the minifilter + kernel callbacks themselves (buildable and testable today in the lab via test-signing, no MVI needed).

Environment audit included (VS 2022 BuildTools + VC++ tools present, but no WDK, no LLVM, test-signing off) so the prerequisites are concrete, not hand-wavy.

Proposes classic C/WDK for the minifilter core rather than `windows-drivers-rs` (Microsoft's own Rust bindings), since that project is explicitly "not yet recommended for production use" and has no minifilter-specific safe wrappers yet — building on it today means raw unsafe FFI with none of Rust'"'"'s safety benefit. Keeps the IPC boundary as the seam: kernel-mode C talks to the existing Rust agent, everything past that boundary (parsing, schema, rules) stays Rust — same pattern already used for `sensor-windows-etw`.

Marked **proposed**, not accepted — this breaks the repo'"'"'s Rust-everywhere convention and should get explicit team sign-off before code lands.

## Test plan

N/A (docs only).


old-dov (approved):
Good scoping ADR — separating the MVI-gated ELAM/PPL piece from the buildable-today minifilter core is the right cut, and the environment audit makes the prerequisites concrete instead of hand-wavy. The `windows-drivers-rs` rejection is well-sourced (checked the project's own README status, not secondhand).

One of your open questions I can partially answer: `agent/src/ipc.rs` doesn't exist, but `crates/ipc` does — already a real precedent, not a blank slate. Per its module doc: it's the agent's local control channel (Windows named pipe / Unix domain socket, peer-authenticated, versioned protocol), currently serving the CLI and `ui/`. It's user-mode-to-user-mode (agent↔CLI), not kernel-mode, so it doesn't answer "does a kernel IPC precedent exist" — the honest answer to that one is still no. But it's worth citing in the ADR's Open Questions/References: whoever implements the minifilter's user-mode side should look at whether the new kernel-driver channel plugs into `crates/ipc`'s existing protocol/auth model rather than inventing a second, parallel IPC surface with its own peer-auth story.

Not a blocker — proposed status is right, this is exactly the kind of detail that should get nailed down in the implementation PR, not gate the ADR. LGTM.

SherlockOmss (approved):
## Review: ADR-0012 Windows Driver Framework

**Status:** ✅ APPROVED

### Summary
Well-scoped ADR that makes a pragmatic decision: separate the ELAM/PPL signing requirements (MVI-gated) from the buildable-today minifilter core (test-signing). The C/WDK choice over windows-drivers-rs is well-justified.

### Key Points

**Scoping Decision ✓**
- Separates ELAM/attestation (MVI required) from minifilter core (buildable now)
- Allows incremental development with test-signing
- Clear path to production signing later

**Technology Choice ✓**
- C/WDK over windows-drivers-rs: Correct decision
- windows-drivers-rs explicitly "not yet recommended for production" (per Microsoft)
- No minifilter-specific safe wrappers available
- Current state would be raw unsafe FFI with no Rust safety benefit

**Architecture Pattern ✓**
- IPC boundary as seam: kernel-mode C ↔ user-mode Rust agent
- Matches existing sensor-windows-etw pattern
- Parsing/schema/rules stay in Rust

**Documentation Quality ✓**
- Environment audit included (concrete prerequisites)
- Open questions documented
- Clear rationale for breaking Rust-everywhere convention

### Note on Open Question
old-dov's comment about `crates/ipc` precedent is valuable - the kernel-driver channel should consider integrating with the existing IPC protocol/auth model rather than creating a parallel surface.

### Verdict
This is exactly the level of detail needed in an ADR. The decision to break the Rust-everywhere convention is justified by practical constraints (windows-drivers-rs maturity). Marked as "proposed" appropriately - needs team sign-off before implementation.

**Recommendation:** Merge as-is. Implementation PR can address the IPC integration question.

Issue #94, opened 2026-09-03
Build sensor-windows-eventlog: WEL channel subscriptions


florianamette:
**Component:** `crates/sensors/windows/eventlog` · see crate doc.

EvtSubscribe push on an allowlist (Security 4624/4625/4648/4672, System 7045, AppLocker/WDAC, Defender operational, Task-Scheduler), rendered fields normalized into schema events (shares the auth/session event type with linux/journal). Driverless; volume-bounded.

## Done when
- [ ] Logon + service-install events from a lab Windows VM normalized end to end
- [ ] Channel allowlist configurable via policy; volume counters in place

SoFrexs:
﻿Progress on #94 from `feature/windows-persistence-detection` (PR to follow) - not closing, this only covers part of the scope.

**Landed:** `crates/sensors/windows/eventlog` polls System 7045 (service install, T1543.003) and Security 4698 (scheduled task, T1053.005) via `wevtutil`, normalized into `schema::Event`. Both validated end-to-end on a real Windows 11 lab VM today.

**Deviation:** polling instead of `EvtSubscribe` push - during the original investigation `EvtSubscribe` proved unreliable for this (Security log not subscribable via ad-hoc ETW, SCM is a legacy provider unsupported by TDH). Details in `docs/adr/0004-windows-persistence-detection-via-eventlog-polling.md`. The native `EvtSubscribe` Win32 API itself was never separately tested - flagged as open in the ADR.

**Still missing:** logon events (4624/4625/4648/4672), AppLocker/WDAC, Defender operational, broader Task-Scheduler coverage, policy-configurable allowlist, volume counters. Can split into follow-up issues if useful.

SoFrexs:
Merged into origin/main.

PR #237, opened 2026-09-18
docs(schema): codify the SCHEMA_VERSION bump contract (closes #114)

### LINKED ISSUE #114: Schema: clarify the version-bump contract for additive optional fields
Review finding on #51/#66: SCHEMA_VERSION stayed 1 while additive optional fields (lineage, enrichment, detection) landed — the docs read as if ANY field change requires a bump. Decide and codify:
- Option A (recommended): amend the contract text — additive, `#[serde(default, skip_serializing_if)]` optional fields are non-breaking and do NOT bump the version; removals/renames/type changes do. Golden fixtures are superseded (never edited) either way.
- Option B: bump to v2 and treat every field addition as a version event (heavy for consumers).

Whichever is chosen, write it into crates/schema/src/lib.rs docs and docs/architecture so the next addition doesn't re-litigate it.

Jihair54:
## Summary

Issue #114: the crate docs said "any change visible in serialization bumps `SCHEMA_VERSION`" without stating what counts as visible, and the issue's originally-recommended Option A ("additive optional fields don't bump") never actually matched real practice.

Checked the real history instead of relitigating A vs B:
- #169 bumped v9 → v10 for a purely additive optional field (`EventMeta::container`).
- #94 bumped v12 → v13 for a new `Event` variant (`Auth`).
- Twelve real bumps (v2 through v13) all follow the same rule — the project already settled on "every serialization-visible change bumps", it was just never written down.

This PR documents what's already been consistently practiced, per old-dov's comment on the issue distinguishing three cases (field / variant / removal-rename-type-change), all landing on the same "bump" outcome:

- `crates/schema/src/lib.rs`: expanded crate-level doc with the three cases, the rationale (non-Rust consumers keying on the exact field set or the `"type"` tag still see a new shape, even when it's non-breaking for a Rust reader), and two explicit non-examples: reusing bits in an existing field (`FLAG_PERSISTENCE_ARTIFACT`, no bump) and a sensor's own unrelated wire ABI tripwire (`sensor-linux-wire::WIRE_VERSION`, bumped for #204 without touching this crate).
- `docs/architecture/event-schema.md`: was a 4-line stub — fleshed out with the full versioning contract and the same worked examples.

## Test plan

- [x] `cargo test -p schema` — 23 golden + 2 v1-compat tests pass (docs-only change, no behavior touched).
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p schema --no-deps` — clean, confirms every intra-doc link (`[EventMeta::container]`, `[FLAG_PERSISTENCE_ARTIFACT]`, etc.) resolves.
- [x] `cargo clippy -p schema --all-targets -- -D warnings` — clean.

Closes #114.


old-dov (approved):
Docs-only, and the content matches reality — spot-checked the FLAG_PERSISTENCE_ARTIFACT and WIRE_VERSION references, both are real and used exactly as described. Good catch codifying "bump on every serialization-visible change, no additive exception" as what the project actually did across 12 real bumps rather than relitigating the issue's original A/B framing. The non-example (bit-flag reuse) and the wire-ABI distinction are the right things to call out — those are the two cases someone will get wrong first. LGTM.

PR #178, opened 2026-09-14
sensor-linux-uprobes: complete implementation (issue #90, phases 1-9)

SherlockOmss:
## Summary

Complete implementation of issue #90 (TLS plaintext taps + shell readline uprobes) through Phase 10.

📖 **Full documentation:** See [`crates/sensors/linux/uprobes/docs/issue-90.md`](https://github.com/synthaea-lab/edr-new/blob/90-build-sensor-linux-uprobes-tls-plaintext-taps-+-shell-readline/crates/sensors/linux/uprobes/docs/issue-90.md) for comprehensive phase-by-phase details.

**Phases 1-6** (Foundation): Symbol resolution, wire structs, eBPF uprobes, userspace loader, schema extension, configuration + budget enforcement.

**Phase 7** (Critical Fixes): Fixed eBPF verifier risk (batch read) + lib_type detection (separate probe functions per library).

**Phase 9** (Security Enhancements): Field-level redaction, data flow documentation, compliance modes (GDPR/HIPAA/PCI-DSS).

**Phase 10** (Operator Guide): Comprehensive deployment guide with configuration examples, security best practices, and future CLI/YAML design.

---
Complete implementation of issue #90 (TLS plaintext taps + shell readline uprobes) through Phase 10:

**Phases 1-6** (Foundation): Symbol resolution, wire structs, eBPF uprobes, userspace loader, schema extension, configuration + budget enforcement.

**Phase 7** (Critical Fixes): Fixed eBPF verifier risk (batch read) + lib_type detection (separate probe functions per library).

**Phase 9** (Security Enhancements): Field-level redaction, data flow documentation, compliance modes (GDPR/HIPAA/PCI-DSS).

**Phase 10** (Operator Guide - NEW ✅): Comprehensive deployment guide with configuration examples, security best practices, and future CLI/YAML design.

---

## Phase 1 - Symbol resolution spike (commit cf0541c)
- Symbol resolution using goblin ELF parser
- Scans system for SSL libraries (OpenSSL, GnuTLS)
- Resolves `SSL_read`/`SSL_write` offsets from dynamic symbol tables
- Validated on 18 libraries, 40 symbols resolved successfully

## Phase 2 - Wire structs definition (commit 259ac75)
- Added `TlsCaptureEvent` (256 bytes budget for TLS plaintext capture)
- Added `ReadlineInputEvent` (512 bytes budget for shell command capture)
- Bumped `WIRE_VERSION` from 3 to 4
[...]

PR #238, opened 2026-09-18
sensor-linux-journal: auth events (sshd/sudo/su) + agent wiring (issue #93)

old-dov:
## Summary

Continues #93's foundation (classify/parsing, merged in #175) with the part it deliberately deferred: mapping into the shared `schema::AuthEvent` (issue #94/ADR-0005 landed in the meantime, so the cross-platform blocker that foundation PR was waiting on is gone) and wiring the journal tail into the running agent.

- `crates/sensors/linux/journal/src/auth.rs` (new): `to_auth_event(record, event)` maps `sshd` accept/fail → `Logon`/`LogonFailure`, `sudo` → `PrivilegedSession` (target `root`) or `ExplicitCredentials` (target anyone else, per `AuthKind::ExplicitCredentials`'s own doc), `su`/local `login` PAM sessions → `ExplicitCredentials`/`Logon`. `sshd`/`sudo` PAM-session lines are deliberately *not* separately mapped — same real action as the accept/sudo-command line already covered, mapping both would double-count.
- `record.rs`: added `_PID`/`_GID` (journald's own trusted fields) to fill `EventMeta.pid`/`User::Unix`.
- `agent/src/commands/linux.rs`: `spawn_journal_tail` — same "poll/tail source with no `Sensor` impl, caller owns the handoff" shape as the existing netlink poller (#92). Best-effort: no `journalctl` (Alpine/OpenRC) logs once and skips, doesn't fail `agent run`.

## Real-hardware validation (Hyper-V lab, Arch Linux `7.2.4-arch1-2`)

- SSH login (`Accepted publickey`) → normalized `AuthEvent{outcome: success, kind: logon, target_user, source_address}`. ✅
- 3-attempt failed-login burst (nonexistent users) → 3× `AuthEvent{outcome: failure, kind: logon_failure}`, correct `target_user`/`source_address` each. ✅
- `sudo <cmd>` → `AuthEvent{kind: privileged_session, target_user: "root"}`. ✅

**Found and fixed along the way:** this box's OpenSSH (9.8+ privsep refactor) re-execs the per-connection worker as `sshd-session` — every `Accepted`/`Invalid user`/PAM-session line carried `SYSLOG_IDENTIFIER=sshd-session`, never `sshd`. The original classifier (`identifier == "sshd"`, written against documentation — no real `sshd` on the original dev box to check against) would have silently never matched on this real, current OpenSSH. Both identifiers are now checked, with a test case and a doc note. Also confirmed: unprivileged `journalctl` cannot read these records at all on this box (journald's per-unit read ACL) — a non-issue in practice since the agent already needs root for eBPF, noted in the crate doc for anyone testing this crate standalone.

## Deliberately out of scope

- Unit lifecycle (`JournalEvent::UnitStarted`/`Stopped`/`Failed`, issue #93's second `Done when` item) has no `schema::Event` mapping here. Inventing a Linux-only "service lifecycle" shape would preempt the same cross-platform decision `AuthEvent` needed (ADR-0005) — Nikolas is actively working the Windows side of the same problem (#224, service-install/lifecycle event reconciliation). Flagged rather than guessed at unilaterally.
- Cursor persistence across agent restarts (`crates/store` integration) — not required by #93's `Done when`; a restart re-tails from "now" rather than resuming, an accepted gap for a follow-up.
- `su` mapping is implemented and unit-tested but not verified against a real interactive `su` prompt (same limitation as the original foundation PR).

## Test plan

- [x] `cargo test -p sensor-linux-journal` — 35 passed (WSL).
- [x] `cargo clippy -p sensor-linux-journal --all-targets -- -D warnings` — clean.
- [x] `cargo clippy -p agent --all-targets -- -D warnings` — clean.
- [x] `tools/check-deps.py` — 38 crates OK.
- [x] Real end-to-end validation on the Hyper-V Arch VM (see above).

Closes the first `Done when` item of #93; the second is intentionally left open per the "deliberately out of scope" section above.


Sollykhan (approved):
LGTM.

Deux points au passage :

**Finding OpenSSH sshd-session** : bien vu — c'est exactement la famille de bug qu'a trouvée Jihair sur #178 (OpenSSL classic API vs _ex, readline dans bash vs libreadline.so). Pattern : *"la doc dit X" n'est pas "vérifié sur X"*. Tu as raison de checker les deux identifiers avec test dédié plutôt que patcher au fil des distros. Ta pratique real-hardware Hyper-V consolide vraiment une convention équipe utile (moi j'ai enchaîné hier avec la baseline capture sur ma machine Windows, et je me monte Hyper-V la semaine prochaine pour #224 justement).

**Ping cross-platform sur unit lifecycle** : confirmé, je bosse sur #224 (réconciliation fixtures XML 4697/7045/4720/4624-4625-4648 contre wevtutil réel sur VM Hyper-V Windows). ETA fin de semaine prochaine. Une fois que j'aurai la vue réelle du côté Windows lifecycle (surtout la distinction install-time vs lifecycle-change du service), on pourra ouvrir un mini-ADR croisé style ADR-0005 pour figer le mapping schema — probablement `LifecycleEvent{outcome, kind: Installed | Started | Stopped | Failed, target_service, user}`. Ton `UnitStarted/Stopped/Failed` sera un candidat direct dans ce schéma. Je te ping quand j'ai avancé.

Sur le reste (cursor persistence deferred, su non testé interactif) : accepté, cohérent avec le "Done when" de #93.

old-dov:
Merci pour le retour détaillé. D'accord sur le plan #224 → mini-ADR croisé pour `LifecycleEvent`, `UnitStarted`/`Stopped`/`Failed` se glissera dedans sans souci vu que c'est déjà le même genre de shape. Je laisse #93 comme ça en attendant, pas de blocage de notre côté.

Issue #133, opened 2026-09-04
Build remediation: rollback and recovery after a response action


florianamette:
Response (#25) kills, quarantines, and isolates — the containment half. The **recovery half is unowned**.

**Goal:** every destructive response is reversible and every cleanup is scoped.
- Restore a quarantined file when a verdict is overturned.
- Remove attacker persistence: services, run keys, scheduled tasks, cron, launchd/systemd units.
- Where the platform supports it, roll back changes (VSS on Windows, snapshots).
- Scope every cleanup by the case's damage manifest — the ransomware pack (#82) already records the files touched in the window; remediation consumes that.

Policy-gated and audited exactly like response itself.

florianamette:
Dependency note: depends on #131 — rollback/recovery is keyed to response actions, which fire on fused verdicts. The verdict model shapes what a remediation record must capture.

Issue #131, opened 2026-09-04
Build verdict fusion: combine engine outputs into one scored verdict


florianamette:
The architecture's `engines → correlator → verdict → response` node, currently unowned. Rules, Sigma, YARA, ML/T0, intel, and the correlator/T1 each emit detections; nothing combines them into a single per-entity verdict with a unified severity/score that `response` gates on.

**Goal:** one verdict per entity/case that response (#25), the sinks, and the case store consume — so response has one thing to gate on and the analyst sees one finding per incident, not N duplicates.

**Needs**
- Cross-engine aggregation keyed by entity/case (the correlator's (ppid, comm) keys are the natural join).
- Deduplication — the same behavior flagged by rules *and* Sigma is one finding.
- A global suppression layer above the per-engine dated FP exclusions.
- Severity/score arbitration: how a high-confidence signature composes with a medium behavioral score into one ranked verdict.

Pairs with the ML T2 scorer and with detection-as-code (#73) for the metadata.

PR #223, opened 2026-09-17
Lab: Windows provisioning scripts (toolchain + agent-install) (#22)

SoFrexs:
﻿## Summary
- Cherry-picks the #22 Windows lab provisioning work onto current `main`: `lab/provisioning/windows-toolchain.ps1` (rustup stable-msvc + Visual Studio Build Tools/Windows 11 SDK for the ETW/Event Log/SCM sensors' MSVC linker) and `lab/provisioning/agent-install.ps1` (stages a built agent + rules onto a machine, optional watchdog service registration).
- Adds the Windows Vagrant usage section to `lab/vagrant/README.md`, reconciled with the Alpine/VirtualBox host-setup section already on `main` (from #186) — same file, non-overlapping content, merged cleanly.
- Second commit fixes PowerShell 5.1 encoding: Windows PowerShell 5.1 misread em-dashes in the scripts as mojibake, switched to ASCII-only.

## Test plan
- [x] Cherry-picked cleanly except one doc merge (`lab/vagrant/README.md`), resolved by hand — no Rust code touched, so no `cargo build`/`cargo test` needed.
- [ ] Manual: bring up a `win11` Vagrant box (VirtualBox or Hyper-V) and run `windows-toolchain.ps1` + `agent-install.ps1` end to end — not yet done on a real x86 host in this pass.
EOF

PR #239, opened 2026-09-18
sensor-linux-lsm: real-hardware io_uring validation + verifier fix (issue #91)

### LINKED ISSUE #91: Build sensor-linux-lsm: BPF-LSM observation + inline blocking

old-dov:
## Summary

Closes #91's first `Done when` item with a real-hardware validation, and fixes a genuine bug in the already-merged `file_open` LSM program that the validation exposed.

- `agent/tests/lsm_io_uring_evasion.rs` (new): loads the eBPF object, attaches the `file_open` LSM hook, drives a real `fio --ioengine=io_uring` write, asserts the hit counter increased. Self-skips (not a failure) without root, without `fio`, or on a kernel without BPF-LSM support — same discipline as `sensor-linux-journal`'s `journalctl`-dependent test.
- `crates/sensors/linux/ebpf/src/main.rs`: fixes the `file_open` program's retval passthrough (see "Bug found" below).
- `sensor-linux-lsm/Cargo.toml`: no changes to its runtime deps — it stays schema-only per `tools/check-deps.py`'s "sensor crates depend only on `schema`" rule. The test needed both `sensor-linux` (to load the eBPF object) and `sensor-linux-lsm` (to attach the hook), so it lives in `agent` instead (already allowed to depend on every sensor crate), with `sensor-linux-lsm` added there as a **dev-only** dependency since it has no production caller yet.

## Bug found and fixed

`try_file_open`'s original code deferred to an earlier LSM's own deny:
```rust
let retval: i32 = ctx.arg(1);
if retval != 0 {
    return Ok(retval);
}
```
On real hardware (Hyper-V lab, Arch `7.2.4-arch1-2`, first time this program ever loaded on a real BPF-LSM-enabled kernel — WSL2's BTF-presence check couldn't catch this), the kernel verifier rejected the whole program:
```
At program exit the register R0 has smin=1 smax=4294967295 should have been in [-4095, 0]
```
`arg()`'s extraction loses enough sign-tracking that the verifier can't prove an arbitrary forwarded value satisfies the LSM exit-state contract. Two follow-up attempts at an explicit bounds check — `retval.clamp(-4095, 0)`, then a hand-written `if/else-if/else` — both still failed, because both compile to a branchless sign-extend-and-mask select the verifier's range tracking can't see through (confirmed via the verifier's own disassembly+trace at each attempt).

Fixed by recognizing the actual requirement is narrower than "forward the exact value": this hook only needs to never override an earlier LSM's deny with its own allow. Returning a fixed constant (`Ok(-1)`) when `retval != 0` preserves that property and is trivially verifier-safe (a compile-time constant has a known range). Full blow-by-blow in the commit message and the crate doc.

## What's still open, and why

- **Second `Done when` item** (policy-gated block denies a test exec with a detection recording the denial) — still blocked on `crates/response`/`crates/policy`'s verdict model (#131, #133), which don't exist yet. Wiring blocking ahead of that would mean inventing the verdict contract here, unreviewed.
- **Third `Done when` item** (capability reporting correct on a kernel without `lsm=bpf`) — two gaps: (1) `schema::sensor::Capabilities` has no field for a supplementary poll-based sensor's coverage (same "no `Sensor` impl" shape as `sensor-linux-netlink`/`sensor-linux-journal`) — adding one to a struct shared with Windows/macOS is a cross-cutting decision, not mine to make solo; (2) this crate has no production caller in `agent` yet, so there's no live report to validate in the first place.

## Test plan

- [x] `cargo test -p sensor-linux-lsm` / `-p agent` — green (WSL, self-skip path).
- [x] `cargo clippy -p sensor-linux-lsm --all-targets -- -D warnings` / `-p agent` — clean.
- [x] `tools/check-deps.py` — 38 crates OK (confirms the dev-only dependency placement respects the sensor-crates-depend-only-on-schema rule).
- [x] Real end-to-end validation on the Hyper-V Arch VM: verifier rejection reproduced, fix confirmed, `fio --ioengine=io_uring` write observed by the hook (hit counter incremented, 0 → 1).


Sollykhan (approved):
LGTM sur le fond.

**Le finding et le fix** : bien vu — c'est du niveau senior. Le débug verifier (documenter les 2 tentatives échouées `clamp(-4095, 0)` puis if/else-if/else avec le raisonnement branchless-select) rend le fix compréhensible et défendable, plutôt que "j'ai bricolé jusqu'à ce que ça passe". Ça a une valeur pédagogique pour la prochaine fois que quelqu'un touchera à un hook LSM.

Même famille de finding que #178 Jihair (validation runtime révèle ce que le compile-clean cache) et le sshd-session de ta #238 : *"ça passe cargo check" ≠ "ça marche sur un vrai kernel"*. Le point que Hugo a soulevé sur ma PR #223 la même journée (validation manuelle non faite) rentre dans la même catégorie. On construit vraiment une convention équipe autour de real-hardware validation, c'est appréciable.

**Une question de doc, pas un blocker** : le fix `Ok(-1)` (= `-EPERM`) préserve "never override an earlier LSM's deny with allow", mais perd le code errno spécifique si un LSM antérieur a retourné autre chose que `EPERM` (par ex. `EACCES` = 13 sur un DAC deny, `EPIPE` sur un cas plus exotique). Un opérateur qui debug plus tard "pourquoi ce process a été deny" va voir `EPERM` uniformément côté userspace au lieu du code d'origine. Sur un hook observation-only avec no policy-enforcement, c'est acceptable — mais est-ce documenté quelque part que le retval du sensor-linux-lsm est un "generic deny signal, pas le code d'origine du LSM antérieur" ? Utile de le noter dans la doc du crate ou en commentaire sur `try_file_open`, pour que la personne qui reprendra ce code dans 6 mois ne s'attende pas à un forward transparent.

Sur les deux "still open" (policy-gated block bloqué sur #131/#133, capability reporting bloqué sur le cross-cutting `Capabilities` field) : accepté, ce sont vraiment des dépendances externes, pas de l'évitement.

old-dov:
Bonne remarque, ajouté (`a6bfd49`) : commentaire sur le `return Ok(-1)` qui précise explicitement que c'est un signal de deny générique, pas un forward transparent de l'errno d'origine, avec un avertissement pour qui voudra brancher du blocking réel dessus plus tard (#131/#133).

Issue #103, opened 2026-09-03
Watchdog: tamper resistance for install artifacts and agent binary


florianamette:
**Component:** `watchdog/` · install/uninstall arms + service definitions.

Kill resistance today is two process layers (service manager ⇄ watchdog ⇄ agent), but the **install artifacts themselves** are undefended. An attacker with local admin/root can neutralise the agent without ever killing a process: edit the systemd unit / launchd plist / service binPath, replace the agent binary on disk, delete the alerts file, or `systemctl disable`. Tamper resistance is about raising the cost of that.

Harden the on-disk and service surface, per platform, without pretending we can stop root outright (we can't — the goal is detection + friction + audit trail):

- **File ownership/perms:** install paths, service definitions, and the agent/watchdog binaries owned by root/SYSTEM, `0644`/`0600` as appropriate, world-non-writable. Refuse to run from a world-writable directory.
- **Config/binary integrity:** the watchdog verifies the agent binary (and its own service definition) against a known hash/signature before (re)spawn, and alerts on mismatch instead of silently launching a swapped binary.
- **Definition drift detection:** watchdog notices if its service definition was altered or the service was disabled, and reports it (self-protection event) rather than accepting it silently.
- **Windows:** evaluate a protected-process / service ACL hardening (`sc sdset`) so non-SYSTEM callers can't `sc stop`/`delete`.
- **macOS:** scope a proper path forward — a launchd `KeepAlive` daemon is trivially unloadable by root; real tamper resistance wants an Endpoint Security **system extension** (managed, harder to unload). Capture that as the long-term target here or split it out.

This is defensive self-protection for a legitimately-installed security agent (admin-consented), not stealth — it stays visible and auditable. Stealth/hiding is explored separately (see #104).

## Done when
- [x] Install artifacts are root-owned (chown / ownership) and world-non-writable — mode bits via `watchdog/src/tamper.rs::harden_permissions`/`refuse_world_writable_dir`; ownership via `harden_ownership` (PR #259), wired in `service/linux.rs` (Linux only)
- [x] Agent binary integrity is verified before spawn; mismatch raises a self-protection event instead of launching — `watchdog/src/tamper.rs::BinaryPin`, wired in `supervise.rs:194` (Linux; other platforms not yet wired)
- [x] Service-definition tamper / disable is detected and reported — `service::snapshot_definition`/`drift_report`, wired in `supervise.rs:208-219` (Linux only)
[...]

PR #233, opened 2026-09-18
feat(packaging): Linux packaging (deb/rpm + systemd) for issue #36

SherlockOmss:
## Summary

Implements native Linux packaging (.deb and .rpm) with full systemd integration, following the Filesystem Hierarchy Standard (FHS) and the "bootstrap model" architecture.

**Closes #36**

### Architecture: Bootstrap Model

- **Package-owned infrastructure**: Binaries installed to `/var/lib/synthaea/bootstrap/`
- **Updater-managed payload**: `/var/lib/synthaea/current/` symlink points to active version
- **Atomic updates**: Updater switches versions via symlink, no package manager conflicts

### Key Features

**FHS-Compliant Layout:**
- `/usr/bin/synthaea-ctl` - CLI symlink to current version
- `/var/lib/synthaea/bootstrap/` - Package-installed binaries (agent, watchdog, cli)
- `/var/lib/synthaea/current/` - Updater-managed symlink (initially → bootstrap)
- `/var/lib/synthaea/versions/` - Updater version directory
- `/var/log/synthaea/` - Agent logs
- `/etc/synthaea/` - Configuration directory (template for now)
- `/run/synthaea/` - Runtime state (tmpfs, cleared on reboot)

**systemd Integration:**
- Service unit with security hardening (NoNewPrivileges, ProtectSystem=strict)
- ExecStartPre hooks for automated symlink creation
- sysusers.d for dedicated system user creation
- tmpfiles.d for runtime directory management

**Package Support:**
- **.deb**: cargo-deb with metadata in `watchdog/Cargo.toml`
- **.rpm**: Custom spec file with scriptlets (%pre, %post, %preun, %postun)

**Code Changes:**
- `watchdog/src/service/linux.rs` - Package detection (prefers installed units over runtime generation)
- `watchdog/src/paths.rs` - FHS-aware path resolution functions
- Backward compatibility: Development mode (manual installs) still supported

### Files Changed

- **New files** (11): systemd units, sysusers/tmpfiles configs, build scripts, spec files, maintainer scripts, README
- **Modified** (4): Cargo.toml (workspace + watchdog), paths.rs, service/linux.rs
- **Deleted** (3): Cleanup of temporary session and personal files

```
 18 files changed, 680 insertions(+), 1075 deletions(-)
```

### Tested Locally
- [x] Release build (agent: 25MB, watchdog: 1.3MB, cli: 421KB)
- [x] .deb package generation (5.6 MB)
- [x] Package contents verification (ar + tar inspection)
[...]

PR #242, opened 2026-09-18
feat(watchdog): tamper resistance for install artifacts and agent binary (#103)

Jihair54:
## Summary

Kill resistance (`supervise`/`service`) assumes an attacker who kills a process. Nothing defended the install surface itself: an attacker with local root could neutralise the agent without ever killing anything — edit the service definition, swap the agent binary on disk, or drop the install directory's permissions to plant a replacement later.

**Linux only** — Windows service-ACL hardening and macOS's Endpoint Security system-extension path are real items in issue #103's "Done when" list, but not implemented here: no Windows/macOS environment to validate either against in this session. This PR covers the Linux subset only; the Windows/macOS work stays tracked on the parent issue, number 103.

- `watchdog/src/tamper.rs` (new): SHA-256 file hashing, `BinaryPin` (pin a binary's hash once, verify later), `refuse_world_writable_dir`, `harden_permissions`. `BinaryPin`/hashing itself carries no `cfg` gate — hashing and comparing bytes means the same thing on every platform, only Linux wires it up so far.
- `watchdog/src/service/linux.rs`: `resolve_paths` now refuses to install pointing at a binary living in a world-writable directory, and hardens the agent/watchdog binaries and the systemd unit file to explicit permissions instead of trusting umask (the OpenRC script was already explicitly `chmod 755`). Added `snapshot_definition`/`drift_report`: a snapshot of the installed definition file's hash and its service-manager enable-symlink (`multi-user.target.wants/...` for systemd, `/etc/runlevels/default/...` for OpenRC), compared later to report content drift or the service having been disabled.
- `watchdog/src/supervise.rs`: `watchdog_loop` pins the agent binary's hash at startup and re-verifies it before every (re)spawn — a binary swapped on disk after supervision began is refused, not launched. On Linux it also re-checks the service-definition snapshot every loop iteration. Both report through `report_self_protection_event`, appending a JSON-Lines line to the alerts file in the same shape a rule/correlator alert uses (`technique="SELF-PROTECTION"`) — no new `schema`/`sinks` dependency for one fixed-shape line.

## Test plan

- [x] `cargo test -p watchdog --release` — 23/23 pass, including new `tamper::tests` (pin/verify against a swapped/deleted file, world-writable-dir refusal, permission hardening) and a JSON-shape test for `report_self_protection_event`.
- [x] `cargo clippy -p watchdog --all-targets -- -D warnings` — clean.
- [x] Real end-to-end validation on the Alpine lab VM (OpenRC), all four scenarios:
  1. **Permissions**: after `install`, the agent/watchdog binaries and `/etc/init.d/synthaea-agent` are `0755 root:root` — confirmed via `ls -la`.
  2. **World-writable install dir refused**: `install --agent-bin` pointed at a binary inside a `chmod 777` directory → refused with a clear message, exit code 1.
  3. **Binary swap while running**: with the service installed and running, atomically replaced (`mv`, matching a real tamper's rename-over-path, not an in-place write that Linux would block with `ETXTBSY`) the agent binary at its supervised path with a different script, then killed the running agent to force a respawn attempt. Result: the swapped binary **never ran** (no side effect from its script), 5 correctly-shaped `SELF-PROTECTION` alerts logged with the expected doubling backoff (5s → 10s → 20s → 40s → 80s), and normal supervision resumed automatically once the original binary was restored.
  4. **Service-definition drift**: appended a line to `/etc/init.d/synthaea-agent` while the watchdog was running → `SELF-PROTECTION: service definition ... was modified after the watchdog started` logged on the next loop iteration (reported, not blocking — the agent still restarts normally).
  5. **Service disabled**: `rc-update del synthaea-agent default` while running → `SELF-PROTECTION: service was disabled (/etc/runlevels/default/synthaea-agent no longer exists)` logged on the next loop iteration.

Remaining Windows/macOS work stays tracked on the parent issue, number 103.


old-dov (approved):
Careful work, and the four real-VM scenarios (permissions, world-writable-dir refusal, live binary swap, definition drift, service-disable) actually exercise the attack surface #103 describes rather than just unit-testing the primitives in isolation.

Traced a couple of things that could've been subtle bugs and weren't:
- `resolve_paths`'s `agent_abs`/`watchdog_abs` are already `.canonicalize()`d before `harden_permissions`/`refuse_world_writable_dir` run on them — so with the `current` → `versions/vX` symlink layout from #233's packaging, the world-writable check correctly lands on the real versioned directory, not the symlink itself.
- `resolve_paths` (and thus the permission/dir hardening) is only reached from `cmd_install`, never from the `run` path — so `watchdog_loop`'s own `BinaryPin` re-verification is the only thing running on every loop iteration, not a permissions rewrite. Matches the module doc's framing (install-surface hardening vs. per-spawn integrity check as two separate concerns).
- `report_self_protection_event`'s hand-escaping only covers `\` and `"`, but every interpolated value here is a fixed install-time path or static message text, never attacker-controlled input reaching this call — a `sinks`-shaped structured writer would be overkill for one fixed-format line, as the comment says.

`BinaryPin::verify()` treating any I/O error (deleted, permission changed) as a mismatch rather than propagating is the right call — "can't prove it's still good" and "proven bad" should get the same refusal.

LGTM — good that the Windows/macOS gap is explicit rather than silently implied-done.

Issue #30, opened 2026-09-03
Build updater: self-update + content distribution client


florianamette:
**Component:** `crates/updater` · **Source:** new development

Staged, signature-verified self-update with rollback; rule/model download via canary rings. Package manager owns bootstrap, updater owns payload (see packaging/README).

## Done when
- [ ] Update applied + rollback on induced failure, in the lab
- [ ] Content update (new rule) reaches a canary agent without binary update

Jihair54:
First implementation slice: **PR #304** — the core \`updater\` library per ADR-0015's Linux-first scope (manifest signing/verification via \`ring\`/Ed25519, the \`bootstrap\`/\`current\`/\`versions\` layout with atomic symlink swap, and local rollback + ban-list bookkeeping). Library-only, VM-validated (34/34 tests, real Ed25519 signing on Linux).

**Not yet done — neither \"Done when\" checkbox is met by this PR alone:**
- ❌ Update applied + rollback on induced failure, in the lab — the mechanism exists and is unit/integration-tested, but there's no actual lab demonstration yet (needs wiring into \`agent\`/\`watchdog\`: heartbeat registration via \`tamper::heartbeat::SilenceMonitor\` for the health check, and a supervised restart — deliberately deferred, since \`updater\` is a leaf crate and can't depend on \`tamper\` directly).
- ❌ Content update (new rule) reaches a canary agent without binary update — canary-ring content distribution (rules/models) is explicitly out of scope for ADR-0015, tracked separately on #73/#49.

Also deferred per the ADR: production key generation/custody (ships test-key-only for now), download transport, Windows/macOS binary self-update.

Next slice: wire this into \`agent\`/\`watchdog\` for the actual lab-validated update+rollback demonstration.

Issue #71, opened 2026-09-03
Build tamper: runtime self-protection (integrity, sensor silence, protected resources)


florianamette:
**Component:** `crates/tamper` · runtime self-protection beyond the watchdog's restart loop. See `docs/architecture/threat-model.md` for the full self-protection map and the honest kernel-adversary ceiling.

Four capabilities:

1. ✅ **Sensor-silence heartbeat** — `tamper::heartbeat` (`SensorHeartbeat` / `SilenceMonitor`): a stopped/detached/frozen sensor stops advancing its counter and the monitor reports it. The platform-agnostic core of audit F-2, shipped.
2. ✅ **Manifest-verified self-integrity** — `tamper::integrity` (`Manifest::verify`): the agent/watchdog binaries and config are hashed against the signed updater manifest; a swap or edit is a high-severity detection. Shipped.
3. ✅ **Protected-resource monitoring** (Linux only, no updater allowlist yet) — watch the agent's own files/services/keys/units through the normal event stream; any non-`updater` writer is a high-severity detection. Consumes the live event stream, so it wires in at the agent (not a standalone primitive).
4. ✅ **Kill-loudness** — spool the termination attempt (who/when/how) before dying, where the platform allows.

**Remaining work**
- [ ] Wire the shipped heartbeat + integrity into the agent/watchdog (pulse each sensor; run the monitor from the watchdog; run `verify()` on a cadence) — needs the updater manifest (#30) for the real root of trust, and a `schema::detection` tamper variant (gated on #114).
- [x] Build protected-resource monitoring. — shipped: agent/src/protected.rs (#244). Remaining gaps: no updater allowlist (blocked on #30), config/service-def/key paths not covered, unlink/rename blind spot, Windows/macOS not wired.
- [x] Build kill-loudness. — shipped: agent/src/kill_loudness.rs

## Done when
- [x] Lab: stopping a sensor without killing the agent produces a local alert within the heartbeat interval — *primitive done; end-to-end wiring pending*
- [x] Lab: modifying the agent binary/config fires the integrity detection on next check — *primitive done; wiring pending*
- [x] Touching agent files from a foreign process fires protected-resource detection — verified on Linux via `agent/src/protected.rs` tests

PR #244, opened 2026-09-18
feat(agent): sensor-silence + protected-resource detection via the live event stream

Jihair54:
## Summary

Wires two of issue, number 71's four capabilities into the Linux agent.

### 1. Sensor-silence detection

Each running sensor (`linux-ebpf`, `linux-netlink`, `linux-journal`) pulses a
`SensorHeartbeat` as it processes events/polls. A dedicated thread polls a
shared `SilenceMonitor` every 10s and turns any confirmed stall into a real
`T1562` alert through `DetectionSink::emit` — an attacker who detaches a
sensor without killing the agent process leaves no crash for the watchdog to
catch; this catches that instead. The existing health-beacon extension point
(`SensorHealthSource`, #134) now gets a real snapshot from the same monitor,
replacing the `NoopSensorHealth` placeholder (removed — no remaining
non-test caller). `linux-netlink` has a genuine self-canary (fixed 10s poll
cadence) so its deadline is tight (30s); `linux-ebpf`/`linux-journal` have
none, so they get a generous 120s deadline relying on ordinary host activity.

### 2. Protected-resource monitoring

`ProtectedResourceGuard` wraps the eBPF-sourced `EventSink` chain (the only
sensor producing `FileOpen` events today). A write-intent open of one of the
agent's own known files — its executable, `alerts.ndjson`, `events.jsonl`,
the `#102` heartbeat file — from any pid other than the agent's own fires a
real `T1562` alert, the same path a rule/correlator/Sigma finding uses.
Promotes `rules::has_write_intent` from `pub(crate)` to `pub` so the agent
reuses the exact same `open(2)` write-intent semantics the persistence rules
already rely on. Path matching tolerates the same known eBPF-collector
limitation `rules::check_persistence_write` documents (a path can be
relative to an unresolved directory fd) via component-suffix matching.

## Out of scope (remaining parts of issue, number 71)

- Self-integrity verification of the agent binary/config — blocked on the
  signed-manifest updater (#30).
- Kill-loudness (spooling a termination attempt before dying) — separate
  capability, not touched here.
- The watchdog binary and the systemd unit/OpenRC install surface are not
  part of protected-resource monitoring's path set — that surface is already
  covered from the install/supervise side by `watchdog::tamper` (#103); naming
  it here would mean guessing the packaging layout or a cross-binary
  dependency.
- True deletion (`unlink`/`rename`) isn't observable — `FileOpenEvent` only
  covers `open(2)`.

This PR does not close issue, number 71.

## Validation

- `cargo test -p agent -p rules -p tamper`: 89/89 passing.
- `cargo clippy -p agent -p rules -p tamper --all-targets -- -D warnings`: clean.
- Real end-to-end runs on the lab VM (not mocked):
  - Ran `agent run` for 150s: `linux-journal` genuinely never pulses on this
    host (journalctl isn't installed on Alpine) — after exactly 120s the
    monitor fired a real `T1562` silence alert into `alerts.ndjson`, while
    `linux-ebpf`/`linux-netlink` (active) produced zero false positives.
  - From a second shell, ran `echo "..." >> alerts.ndjson` against the
    agent's own live alert file. The eBPF sensor caught the real `open(2)`
    from that foreign `sh` process and the agent emitted:
    ```
    {"technique":"T1562","message":"pid 2318 (`sh`) opened protected agent resource alerts_71b.ndjson for write"}
    ```

## Test plan

- [x] Unit tests (agent + rules + tamper)
- [x] Clippy clean
- [x] Real VM run: true-positive silence alert, zero false positives from active sensors
- [x] Real VM run: genuine foreign-process write firing the protected-resource detection


old-dov (approved):
Clean wiring of already-solid primitives (`crates/tamper::heartbeat`, itself well-tested independently — checked its dedup logic: `poll()` correctly reports once per silence episode and re-arms after recovery, no spam risk). Deadlines are sane: 30s (3×10s poll) for the netlink sensor's real canary, 120s generous fallback for ebpf/journal's no-canary case, matching the crate's own documented F-2 limitation honestly rather than pretending a synthetic canary exists where it doesn't.

One granularity note, not a blocker: `spawn_netlink_poller` shares a single heartbeat across both its sub-sources (listen-port + conntrack). If conntrack alone breaks (e.g. the nftables `ct`-rule precondition from #92) while listen-port keeps polling successfully, the netlink sensor as a whole still looks healthy — silence detection is per named sensor, not per capability within it. Consistent with the one-heartbeat-per-sensor design here, and probably fine for v1, but worth knowing if a future incident is "conntrack silently dead but listen-port fine."

Real validation (120s run catching journal's genuine absence with zero false positives from the active sensors) is exactly the right way to prove this doesn't cry wolf on an idle-but-healthy host. LGTM.

old-dov:
Flagging a finding on the last commit (`91ee35b`, protected-resource monitoring) — this landed *after* my approval on the sensor-silence half, so it never actually got reviewed. Not blocking anything since this is already merged, but worth tracking on #71.

`matches_protected`'s dirfd-relative fallback (matching on a single path component when the eBPF collector only captured a bare filename) is the right general shape — component-based suffix matching correctly avoids the raw-substring trap (`"myagent"` vs `"agent"`), same care `check_persistence_write` doesn't need because its patterns (`.bashrc`, `/etc/cron.d/`) are inherently distinctive strings.

But the protected filenames here are not distinctive the same way: `alerts.ndjson`, `events.jsonl`, and especially the agent's own exe basename (`agent`, going by `protected_paths`' `current_exe()`) are generic enough that an unrelated file/process sharing that exact basename elsewhere on the host is plausible — a second agent instance, an unrelated tool's own `events.jsonl`, or literally any binary named `agent`. If the eBPF path capture hits the documented dirfd-relative case for one of those, `matches_protected` has zero directory context left to disambiguate and will alert on something that was never actually the protected resource.

Concretely reproducible with the crate's own test helper: `matches_protected(Path::new("/opt/synthaea/agent"), "agent")` is `true` — and would be exactly as true for any other process anywhere named `agent`.

Not sure the fix — narrowing to "at least 2 matching components" would reduce false positives but might reintroduce misses on the exact dirfd-relative case this was built to tolerate. Flagging so it's a deliberate follow-up decision on #71 rather than a silently-accepted gap.

Issue #25, opened 2026-09-03
Build response: kill, quarantine, isolate


florianamette:
**Component:** `crates/response` · **Source:** new development (audit P9)

First actions: kill process (opt-in, high-confidence correlated verdicts), quarantine written payloads. Host isolation later. Every action policy-gated, audited, reversible where possible (unquarantine).

## Done when
- [ ] Beacon scenario with response enabled: process killed, payload quarantined, both audited
- [ ] Policy off = observe-only, verified by test

PR #248, opened 2026-09-18
feat(response): automated kill + quarantine on a verdict (issue #25)

Jihair54:
Implements the first two automated response actions from #25 — process kill and
payload quarantine — both policy-gated, audited, and reversible where possible.
Host isolation is out of scope for this pass (tracked as the remaining part of
#25, same as #71/#103's Linux-first precedent).

## What's new
- `policy::ResponsePolicy` (`kill_enabled`, `quarantine_enabled`, defaults to
  **both false** — policy off = observe-only, per #25's acceptance criterion).
- `response::kill_process` — policy gate + injected `terminate` closure, so
  `response` stays platform-neutral (`crates/response` is leaf-tier: only
  `schema`+`policy`); the actual `SIGKILL` lives in `agent`.
- `response::quarantine_file`/`unquarantine` — platform-neutral (`std::fs`
  only): moves the payload into `quarantine_dir` under its own SHA-256 hex
  name, marks it read-only, writes a `<hash>.origin` sidecar for reversal.
- `agent`: new `--enable-kill`/`--enable-quarantine` CLI flags (both default
  off — no config-file/ADR mechanism exists yet for this). Kill triggers on a
  correlated BAYES verdict; quarantine triggers on a confirmed YARA match.
  Both actions are audited through the existing `alerts.ndjson` path, tagged
  `RESPONSE-KILL`/`RESPONSE-QUARANTINE`.

## Bug found and fixed during validation
Quarantining a file is itself a write the eBPF sensor observes, which was
re-triggering YARA + a second quarantine attempt on the file's own copy,
corrupting its `.origin` sidecar. Fixed with a `starts_with(quarantine_dir)`
guard.

## Validation (Alpine lab VM, real conditions, not mocked)
- Kill: a real `nc` beacon to an external IP crossed the BAYES threshold and
  was actually `SIGKILL`ed; audited in `alerts.ndjson`.
- Quarantine: a real YARA-marked file was moved into quarantine, hashed,
  made read-only, and its `.origin` sidecar was correct — confirmed quarantined
  exactly once (no self-rescan cascade after the fix above).
- `cargo test`/`cargo clippy -D warnings`/`tools/check-deps.py` all clean.

Part of #25 — host isolation remains open scope, issue stays open.


old-dov (approved):
Reviewed + validated via WSL (cargo test -p response -p policy -p sensor-linux -p agent, 66 tests green, no failures). Design notes: kill/quarantine both default-disabled via ResponsePolicy::default(), terminate_process correctly uses SIGKILL with checked libc::kill return, quarantine_file guards against re-quarantining its own output (real bug found on lab VM per the code comment), unquarantine keeps the file read-only post-restore. LGTM.

Issue #90, opened 2026-09-03
Build sensor-linux-uprobes: TLS plaintext taps + shell readline


florianamette:
**Component:** `crates/sensors/linux/uprobes` (probe programs join the ebpf crate build) · see crate doc.

SSL_read/SSL_write uprobes on OpenSSL/BoringSSL/GnuTLS for pre-encryption C2 content (budgeted: first N bytes, comm allowlist), bash/zsh readline for interactive commands. Symbol resolution per binary, tracked across library updates; schema gains an event type for tapped content (additive).

## Done when
- [ ] Lab: a curl HTTPS beacon's plaintext request line is captured and attributed
- [ ] Interactive shell builtins (no execve) appear as events
- [ ] Budget documented + enforced (bytes/proc/sec caps)

PR #249, opened 2026-09-18
fix(sensor-linux): bump the userspace WIRE_VERSION tripwire to 5

Jihair54:
`main` currently fails to compile `sensor-linux` (and therefore `agent`) for
anyone pulling it — found while validating an unrelated PR (#248) after
merging main.

#178 (issue #90, uprobes) bumped `sensor-linux-wire::WIRE_VERSION` 4 → 5 for
the new `TlsCaptureEvent`/`ReadlineInputEvent` structs, and correctly updated
the new `sensor-linux-uprobes` crate's own copy of the tripwire — but missed
the pre-existing one in `sensor-linux/userspace/src/normalize.rs`, still
pinned at `== 4`.

Audited before bumping: this module only imports `EventMeta`/`ExecEvent`/
`ConnectEvent`/`FileOpenEvent`/`ContainerContext` from the wire crate, none of
which changed shape in v5 — the bump is purely additive from this module's
point of view, so the existing mappings are unaffected.

## Validation (Alpine lab VM)
`cargo test -p sensor-linux -p agent` and
`cargo clippy -p sensor-linux -p agent --all-targets -- -D warnings` both
clean after the fix (both failed before it, with the tripwire's own
assertion panic).


old-dov (approved):
Compile-checked + tested via WSL (`cargo check --workspace`, clean besides pre-existing unrelated `bpf_target_arch` cfg warnings). One-line tripwire bump matching the audit note — LGTM.

Issue #252, opened 2026-09-21
sensor-linux-uprobes: readline_normalizes_bash_input has an off-by-one in its own test data


Jihair54:
## What

\`cargo test -p sensor-linux-uprobes\` panics (doesn't fail an assertion — panics
during setup) on \`normalize::tests::readline_normalizes_bash_input\`:

\`\`\`
thread 'normalize::tests::readline_normalizes_bash_input' panicked at
crates/sensors/linux/uprobes/src/normalize.rs:258:27:
copy_from_slice: source slice length (12) does not match destination slice length (13)
\`\`\`

The test builds a fixed-size event buffer and declares \`input_len: 13\`, then does:

\`\`\`rust
event.input[..13].copy_from_slice(b"ls -la /tmp\n");
\`\`\`

\`b"ls -la /tmp\n"\` is 12 bytes (\`l s space - l a space / t m p \n\`), not 13, so the
slice lengths mismatch and \`copy_from_slice\` panics.

## Why this is separate from #251

Unrelated to \`redact.rs\`/redaction — this is a different module (\`normalize.rs\`)
and a different, pre-existing bug in the test's own literal data, not in the code
under test. Found while validating #251's fix on the lab VM: the crate's test
target only started actually *running to completion* after #250 fixed the compile
error, and this is apparently the first time anyone has seen the full run.

## Fix

Either drop \`input_len\` to \`12\` (matching the literal), or fix the literal to be
13 bytes (e.g. append another char). Whoever picks this up should check the
sibling \`readline_normalizes_zsh_input\` test right below it for the same class of
off-by-one, since \`input_len: 11\` / \`b"export FOO="\` (11 bytes) looks consistent
but is worth double-checking given this pattern already fooled one test.

PR #253, opened 2026-09-21
fix(sensor-linux-uprobes): make TLS/readline redaction actually redact (#251)

### LINKED ISSUE #251: sensor-linux-uprobes: redaction is completely non-functional — every regex pattern fails to compile, raw secrets pass through unredacted
## Summary

The redaction layer for #178's TLS/readline capture (`crates/sensors/linux/uprobes/src/redact.rs`) does not redact anything. Every single pattern in the module fails `regex::Regex::new(...)`, and `redact_pattern`'s defensive fallback silently returns the **original, unredacted** text on any regex error:

```rust
fn redact_pattern(text: &str, pattern: &str, replacement: &str) -> String {
    match regex::Regex::new(pattern) {
        Ok(re) => re.replace_all(text, replacement).into_owned(),
        Err(_) => text.to_string(), // Defensive: invalid regex shouldn't break capture
    }
}
```

Confirmed by running the crate's own test suite (`cargo test -p sensor-linux-uprobes`, blocked until #250 lands — see that PR): **11 of 13 `redact::tests::*` tests fail**, all with the same shape (`assert!(text.contains("... [REDACTED]"))` failing because the text comes back untouched).

This was invisible until now because there's no CI and the crate didn't even compile its test target (unrelated bug, #250) since #178 merged — nobody had run this test suite since it was written.

## Impact

Mitigated by the feature being **disabled by default** (`TlsConfig`/`ReadlineConfig::default().enabled == false`) — no exposure until an operator opts in. But the moment TLS/readline capture is enabled, none of the documented protections apply: decrypted HTTPS bodies and shell input are captured **with live credentials, tokens, and passwords intact**, silently, with no error or log line indicating redaction failed. This directly contradicts `DATA_FLOW.md`/`OPERATOR_GUIDE.md`'s framing of redaction as the safety control that makes capture acceptable.

## Root causes (two independent bugs)

**1. Missing `regex` crate features** — `crates/sensors/linux/uprobes/Cargo.toml`:
```toml
regex = { version = "1.11", default-features = false, features = ["std"] }
```
No `unicode-case`/`unicode-perl`. Every pattern using `(?i)` (case-insensitive) or Unicode-aware `\s`/`\w` fails to compile at runtime (`Regex::new` returns `Err`). This breaks: Authorization header, Cookie header, URL credentials, API keys in query strings, AWS credentials. Verified fixable by adding `"unicode-case", "unicode-perl"` to the feature list — 5 of these patterns pass once added.

**2. Patterns using backreferences, which the `regex` crate cannot support at all** (it's a guaranteed-linear-time, non-backtracking engine — `\1`/`\2` are not valid syntax regardless of feature flags):
- `r#"(?i)\b(export\s+\w+)=(['"]?)([^'"\s]+)\2"#` (export statements)
- `r#"(?i)--password[=\s]+(['"]?)([^'"\s]+)\1"#` (`--password=`)
- `r#"(?i)-p\s*(['"]?)([^'"\s]+)\1"#` (mysql/psql `-p`)
- `r#"-u\s+(['"]?)([^'"\s]+)\1"#` (curl `-u`)

These four will keep failing no matter what features are enabled — they need to be rewritten without backreferences (e.g. two independent capture groups, matching to end-of-token instead of a paired quote), or the crate needs to switch to something that supports backtracking (e.g. `fancy-regex`) for this specific case.

## Suggested fix

1. Quick, safe part: add `"unicode-case", "unicode-perl"` to the `regex` feature list — fixes 5+ of the 11 failing tests immediately.
2. Needs a real decision, not a one-liner: rewrite the 4 backreference-dependent shell-secret patterns. Whoever owns this should also consider making `redact_pattern`'s fallback loud (log/panic in debug, or a startup self-test that verifies every pattern compiles) instead of silently swallowing a `Regex::new` failure — that's what let this ship unnoticed.

Found while validating #248/#249's WSL test run and reported separately per Jean's call, since it's unrelated to the deadline batch and needs an actual redaction-logic decision, not a quick merge.

Jihair54:
## What

Fixes #251: every regex pattern in `redact.rs` failed `Regex::new` at
runtime, and the fallback silently returned the original, unredacted
text — so once TLS/readline capture is enabled, secrets pass through
in the clear with no error or log line anywhere.

## Root causes

1. **Missing regex features.** `regex` was pulled in with
   `default-features = false, features = ["std"]` — no
   `unicode-case`/`unicode-perl`, so every pattern using `(?i)` or
   Unicode-aware `\s`/`\w` failed to compile. Fixed by adding those
   two features in `Cargo.toml`.
2. **Backreferences.** Four patterns (export, `--password`, `-p`,
   curl `-u`) matched a value's closing quote against its opening one
   via `\1`/`\2`. This crate's `regex` engine is guaranteed-linear-time
   and does not support backreferences at all — not a feature-flag
   issue, this syntax is simply invalid for this engine. Rewritten
   with two *independent* optional-quote groups instead; the
   replacement text never echoes the value or its quoting back, so
   this is exact for redaction's purposes.
3. **Found by actually running the fixed patterns** (only possible
   after 1+2 compiled): the new `-p` pattern matched *inside* the
   `--password=[REDACTED]` text the `--password` pattern had just
   produced, because `--password` contains the literal substring
   `-p` (2nd dash + "p"), and greedily consumed through `=`/`[`/`]`
   to clobber it into `--p[REDACTED]`. Fixed by anchoring `-p` on
   "preceded by start-of-string or a non-dash character" (no
   lookbehind in this engine, so the anchor char is consumed and
   echoed back via `$1`).

## Also in this PR

- Hoisted each function's `(pattern, replacement)` pairs into a
  shared `const` list, so the new `all_patterns_compile` regression
  test walks the exact same patterns production code runs — the
  original bug shipped with zero test coverage catching it precisely
  because nothing forced the test and the real call sites to stay in
  sync.
- `redact_pattern`'s fallback now `log::error!`s instead of silently
  swallowing a `Regex::new` failure, so a future regression like this
  one is loud instead of invisible.

## Not in this PR

Filed #252 separately: an unrelated pre-existing off-by-one in
`normalize::tests::readline_normalizes_bash_input`'s own test data,
surfaced while validating this fix on the lab VM. Different module,
different bug, not a redaction issue.

## Validation

On the Alpine lab VM (`cargo test -p sensor-linux-uprobes redact::` and
`cargo clippy -p sensor-linux-uprobes --all-targets -- -D warnings`):
all 14 `redact::tests::*` pass, clippy clean (remaining warnings are
pre-existing, from the `ebpf` build-dependency crate, unrelated to
this change).


old-dov (approved):
Validated via WSL: cargo test -p sensor-linux-uprobes redact:: — all 14 pass (the one remaining failure in the suite, readline_normalizes_bash_input, is the unrelated pre-existing bug correctly split out to #252). cargo clippy -p sensor-linux-uprobes --lib --tests -D warnings clean (only pre-existing unrelated bpf_target_arch cfg warnings from the ebpf build-dep). Good catch on the --password/-p clobbering, and the loud log::error! + all_patterns_compile regression guard are exactly the right response to how #251 shipped unnoticed. LGTM.

SherlockOmss (approved):
## Excellent fix for a critical silent failure

This PR addresses a severe issue where **all redaction patterns were silently failing** at runtime, allowing secrets to pass through in plaintext. The three-root-cause analysis is spot-on, and the fixes are correct and well-documented.

---

## Root Cause Analysis ✓

### 1. Missing regex features (Cargo.toml:24)

**Problem:** `regex = { features = ["std"] }` without `unicode-case`/`unicode-perl` caused every pattern using `(?i)` or `\s`/`\w` to fail compilation.

**Fix:** Added both features:
```toml
regex = { version = "1.11", default-features = false, features = ["std", "unicode-case", "unicode-perl"] }
```

This is correct — the regex crate's Unicode support is feature-gated, and case-insensitive matching + Perl-style character classes require explicit opt-in.

### 2. Backreferences incompatible with DFA engine (redact.rs:65-106)

**Problem:** Four patterns used `\1`/`\2` to match closing quotes against opening ones:
```rust
// Old (invalid):
r#"\b(export\s+\w+)=(['"]?)([^'"\s]+)\2"#  // \2 is a backreference
```

The `regex` crate uses a guaranteed-linear-time DFA engine that **fundamentally cannot support backreferences** (this isn't a feature flag — the syntax is invalid for this engine).

**Fix:** Replaced with independent optional quote groups:
```rust
// New (valid):
r#"\b(export\s+\w+)=(['"]?)([^'"\s]+)(['"]?)"#  // Two independent optional groups
```

**Correctness:** This is safe for redaction because the replacement text (`$1=[REDACTED]`) never echoes the value or its quoting back. An unpaired quote match (e.g., matching `'foo"` instead of `'foo'`) has no observable negative effect — it's best-effort redaction, not a parser.

### 3. `-p` pattern matching inside `--password` (redact.rs:94-98)

**Problem:** The literal substring `-p` appears in `--password` (second dash + "p"). After `--password=foo` is redacted to `--password=[REDACTED]`, an unanchored `-p` pattern greedily matches the `-p` inside it and consumes through `=`/`[`/`]`, producing `--p[REDACTED]`.

**Fix:** Anchored on "not preceded by another dash":
```rust
r#"(?i)(^|[^-])-p\s*(['"]?)([^'"\s]+)(['"]?)"#
//     ^^^^^^^ Anchor: start-of-string or non-dash character
```

The anchor character is consumed into the match and echoed back via `$1`, which is the correct approach for a DFA engine without lookbehind.

**Order dependency:** This only matters because `--password` runs first in the list. The comment at redact.rs:76-82 correctly documents this — excellent defensive annotation.

---

## Code Quality Improvements ✓

### Hoisted patterns to `const` arrays (redact.rs:22-37, 83-106)

**Why this matters:** The original bug shipped with **zero test coverage** because the test fixtures and the production call sites had drifted out of sync. By hoisting to:
```rust
const TLS_PATTERNS: &[(&str, &str)] = &[...];
const READLINE_PATTERNS: &[(&str, &str)] = &[...];
```

...the new `all_patterns_compile` test (redact.rs:250-257) walks the **exact same patterns** production code runs. A future regression can't slip through the same way.

### Loud failure on pattern compilation error (redact.rs:136-143)

**Before:** Silent fallback returning unredacted text.

**After:**
```rust
Err(err) => {
    log::error!("redact_pattern: pattern {pattern:?} failed to compile: {err}");
    text.to_string()
}
```

This is the right call — still fail-open (dropping a capture or crashing the sensor over a redaction bug is worse), but **loud** about it via logs instead of invisible.

---

## Testing ✓

The `all_patterns_compile` test is simple but critical:
```rust
#[test]
fn all_patterns_compile() {
    for (pattern, _) in TLS_PATTERNS.iter().chain(READLINE_PATTERNS.iter()) {
        assert!(
            regex::Regex::new(pattern).is_ok(),
            "pattern failed to compile: {pattern}"
        );
    }
}
```

This would have caught the #251 regression immediately.

**Validation:** All 14 tests pass + clippy clean (per the PR description).

---

## Documentation ✓

The comments at redact.rs:65-82 are exemplary:
- Explain **why** backreferences don't work (DFA engine limitation)
- Explain **why** the independent-quote-group workaround is safe (replacement doesn't echo back)
- Explain **why** `-p` needs the anchor (order dependency with `--password`)
- Note the tradeoff (best-effort redaction, not a parser)

This level of "why" documentation prevents future regressions where someone "simplifies" the pattern without understanding the constraint.

---

### 1. Pattern order matters

The comment correctly notes that `-p` must run after `--password` because the anchor depends on it. This is a fragile invariant (reordering the const array breaks correctness), but acceptable given:
- It's documented
- The alternative (lookahead/lookbehind) isn't available in this engine
- Redaction is best-effort, not security-critical

### 2. Independent quote groups allow mismatched quotes

The new patterns match `export FOO='bar"` (mismatched quotes) whereas the old backreference-based ones wouldn't. But:
- The replacement doesn't echo the value back, so this has no observable effect
- Redaction is best-effort pattern matching, not a shell parser
- This is an acceptable tradeoff for correctness in the regex engine

---

## Verdict

**LGTM — merge immediately.** This is a critical security fix with:
- ✅ Correct root-cause diagnosis (all three issues)
- ✅ Correct fixes (features, no backreferences, anchored `-p`)
- ✅ Excellent documentation (explains constraints and tradeoffs)
- ✅ Regression test (`all_patterns_compile`)
- ✅ Improved observability (log errors instead of silent fail)

**Impact:**
- Fixes #251 (all redaction patterns failing silently)
- Enables TLS/readline capture to actually redact secrets
- Prevents future regressions via test coverage

**Recommendation:** Consider adding a comment in the `all_patterns_compile` test itself referencing issue #251, so future readers understand why this seemingly-trivial test exists (it's guarding against a class of silent failures).

PR #256, opened 2026-09-21
fix(agent): gate the kill_loudness module to Linux, matching its dependency

Jihair54:
## What

\`mod kill_loudness;\` in \`agent/src/main.rs\` had no \`cfg\` gate at all, but the
module calls raw POSIX signal APIs (\`sigemptyset\`/\`pthread_sigmask\`/\`sigwaitinfo\`)
that don't exist in the Windows build of the \`libc\` crate — and
\`agent/Cargo.toml\` already only pulls \`libc\` in under
\`[target.'cfg(target_os = "linux")'.dependencies]\`. Result: \`agent\` has failed
to compile on Windows since kill-loudness (#71) landed, with 14
\`use of unresolved module or unlinked crate \`libc\`\` errors.

## Why nobody caught this

Found while standing up a Windows Rust toolchain for the first time on our dev
box, to validate #103's Windows service ACL hardening (needed a real
\`agent.exe\`). Nobody had built \`agent\` on Windows before — Linux work goes
through the lab VM, and CI is currently disabled repo-wide (billing).

## Fix

\`mod kill_loudness;\` → \`#[cfg(target_os = "linux")] mod kill_loudness;\`,
matching the \`libc\` dependency gate. Its only caller
(\`agent/src/commands/linux.rs\`) is already Linux-only, so nothing else changes.

## Validation

\`cargo build -p agent\` and \`cargo test -p agent\` on Windows (MSVC toolchain,
native host): compiles clean, all 14 tests pass.


old-dov (approved):
Verified on native Windows (not just WSL/emulation): cargo test -p agent — 14/14 pass, agent compiles again. Minimal, correctly-scoped fix (kill_loudness's only caller was already Linux-only).

PR #257, opened 2026-09-21
feat(watchdog): harden the Windows service ACL against non-SYSTEM stop/delete (#103)

Jihair54:
## What

Part of #103: a fresh \`sc create\` gets the stock Windows service DACL, which
lets any local Administrator (not just SYSTEM) \`sc stop\`/\`sc delete\`/\`sc
pause\` the service — an attacker with local admin (but not SYSTEM) can
neutralize the agent with one \`sc stop\` + \`sc delete\`, no process kill
involved.

## Design

\`cmd_install\` now applies a hardened SDDL via \`sc sdset\` after the service is
created and started:
- \`SY\` (LocalSystem) keeps full control.
- Built-in Administrators (\`BA\`) keep query/start/interrogate/read-control/
  write-dac/write-owner, but lose stop/pause/delete/change-config.
- Everyone else (\`IU\`/\`SU\`/\`WD\`) gets read-only rights.

\`BA\` keeping \`WRITE_DAC\` is deliberate: \`cmd_uninstall\` resets the ACL back
to a permissive one before its own stop/delete, so \`watchdog uninstall\` still
works for an admin — only a bare \`sc stop\`/\`sc delete\` typed by hand,
bypassing the tool, is denied. \`cmd_install\` also resets first, so a
re-install/upgrade over an already-hardened service doesn't fail on the \`sc
description\`/\`sc failure\` calls (both need \`SERVICE_CHANGE_CONFIG\`).

This is friction, not a real boundary against local admin — matches the
issue's own framing (detection + friction + audit trail, not a claim to stop
root outright).

## Validation

Standing up a Windows Rust toolchain for the first time on our dev box (see
#256 — a separate, unrelated Windows-build-breaking bug this surfaced along
the way), then real end-to-end against an actually-installed service:

\`\`\`
> sc sdshow SynthaEDR
D:(A;;CCDCLCSWRPWPDTLOCRSDRCWDWO;;;SY)(A;;CCLCSWRPLOCRRCWDWO;;;BA)(A;;CCLCSWLOCRRC;;;IU)(A;;CCLCSWLOCRRC;;;SU)(A;;CCLCSWLOCRRC;;;WD)...

> sc stop SynthaEDR      # run as a local Administrator, not SYSTEM
[SC] OpenService FAILED 5: Access is denied.

> sc delete SynthaEDR
[SC] OpenService FAILED 5: Access is denied.

> watchdog.exe uninstall
[SC] SetServiceObjectSecurity SUCCESS
[SC] DeleteService SUCCESS
[watchdog] service "SynthaEDR" removed.

> sc query SynthaEDR     # a few seconds later
[SC] OpenService FAILED 1060: The specified service does not exist.
```

Plus \`cargo test -p watchdog\` (9/9 passing, including 4 new tests asserting
the SDDL strings themselves grant/deny the right rights) and \`cargo clippy -p
watchdog --all-targets -- -D warnings\` clean.

## Not in this PR

- Ownership (\`chown\`-equivalent) hardening on Linux — still open, tracked in
  #103.
- macOS Endpoint Security path — still open, tracked in #103.


old-dov (approved):
Verified on native Windows: cargo test -p watchdog — 28/28 pass (incl. the 4 new SDDL tests), cargo clippy -p watchdog --all-targets -D warnings clean. Hand-decoded HARDENED_SERVICE_SDDL against the documented rights list — BA correctly loses WP/DT/SD (stop/pause/delete) while keeping WD (needed for cmd_uninstall's own reset-then-delete), SY keeps GA. Good design: friction against non-SYSTEM local admin, explicitly not claimed as a hard boundary, and the tool's own uninstall path still works. LGTM.

PR #259, opened 2026-09-21
feat(watchdog): harden install artifact ownership to root on Linux (#103)

Jihair54:
## What

Part of #103: \`harden_permissions\` only clears bad mode bits — a
non-root-owned binary or service definition is one \`chmod\` away from being
writable by its own (non-root) owner again, which undermines the point of
world-non-writable mode bits.

## Change

Adds \`tamper::harden_ownership\` (\`std::os::unix::fs::chown\` to uid/gid 0)
and wires it in alongside every existing \`harden_permissions\` call:
- the agent and watchdog binaries (\`resolve_paths\`'s loop)
- the systemd unit (\`install_systemd\`)
- the OpenRC script (\`install_openrc\`)

\`install\`/\`uninstall\` already require root to write these paths at all
(\`/etc/systemd/system\`, \`/etc/init.d\`, the resolved binary paths) — this
normalizes ownership on a path the process could already write, not an
escalation.

## Validation

On the Alpine lab VM, as root:

- \`cargo test -p watchdog tamper::\` — 7/7 passing, including a new
  \`harden_ownership_sets_root_when_running_as_root\` test (skips itself with a
  clear message when not run as root).
- \`cargo clippy -p watchdog --all-targets -- -D warnings\` — clean.
- Real end-to-end: chowned the built \`agent\` binary to a throwaway
  non-root user, then \`watchdog install\`:

\`\`\`
$ ls -l target/debug/agent
-rwxr-xr-x 2 testuser testuser 311224520 ... target/debug/agent

$ ./target/debug/watchdog install --agent-bin ./target/debug/agent
[watchdog] OpenRC service installed and started.

$ ls -l target/debug/agent /etc/init.d/synthaea-agent
-rwxr-xr-x 2 root root 311224520 ... target/debug/agent
-rwxr-xr-x 1 root root       464 ... /etc/init.d/synthaea-agent

$ ./target/debug/watchdog uninstall
[watchdog] synthaea-agent OpenRC service uninstalled.
```

Both the binary (previously non-root-owned) and the generated OpenRC script
came out \`root:root\`.

## Not in this PR

macOS Endpoint Security path — still open, tracked in #103.

### CODE REVIEW

**Review by SherlockOmss — APPROVED** [review]
## Review: Watchdog Ownership Hardening (#103)

**Status:** ✅ APPROVED

### Summary
Excellent security hardening that closes a real gap: mode bits alone don't prevent a non-root owner from making their own binary writable again. The fix is simple, correct, and well-tested.

---

## Problem Analysis ✓

**Gap identified:**
```rust
// Before: only harden_permissions
harden_permissions(bin, 0o755)  // Sets mode to rwxr-xr-x

// Problem: if owner != root
$ ls -l agent
-rwxr-xr-x 1 testuser testuser ... agent  // testuser can chmod again!

$ chmod u+w agent  // Non-root owner can make it writable
-rwxr-xr-x 1 testuser testuser ... agent
```

**Impact:** Undermines the entire point of world-non-writable mode bits if the owner can just change them back.

---

## Solution ✓

**Added `harden_ownership`:**
```rust
pub(crate) fn harden_ownership(path: &Path) -> std::io::Result<()> {
    std::os::unix::fs::chown(path, Some(0), Some(0))
}
```

**Applied everywhere permissions are hardened:**
1. Agent/watchdog binaries (`resolve_paths` loop)
2. Systemd unit (`install_systemd`)
3. OpenRC script (`install_openrc`)

**Correct pairing:**
- `harden_permissions` → sets mode bits (0o755 or 0o644)
- `harden_ownership` → sets owner to root:root
- Together: Complete protection

---

## Security Assessment ✓

**Not an escalation:**
- `install`/`uninstall` already require root
- These paths (`/etc/systemd/system`, `/etc/init.d`, binary locations) already need root to write
- This normalizes ownership on paths the process could already modify
- Documented clearly in function comment

**Defense in depth:**
- Mode bits: Prevent world/group writes
- Ownership: Prevent owner bypass
- Both required for complete protection

---

## Code Quality ✓

**Documentation excellent:**
```rust
/// Sets `path`'s owner and group to root (uid/gid 0) — mode bits alone
/// (`harden_permissions`) still leave a non-root-owned install artifact one
/// `chmod` away from being writable by its own (non-root) owner again.
```

Explains the "why", not just the "what". Future maintainers won't wonder why both functions exist.

**Error handling:**
```rust
crate::tamper::harden_ownership(bin)
    .with_context(|| format!("hardening ownership on {}", bin.display()))?;
```

Proper error propagation with context - matches existing pattern.

---

## Testing ✓

### Unit Test
```rust
#[test]
fn harden_ownership_sets_root_when_running_as_root() {
    if unsafe { libc::geteuid() } != 0 {
        eprintln!("skipping: chown(2) to root requires root...");
        return;
    }
    // ... actual test
}
```

**Excellent graceful skip pattern:**
- Doesn't fail on non-root CI
- Clear message explaining why it skipped
- Actually tests when possible (root environments)

### Real End-to-End Validation
```bash
# Before install
-rwxr-xr-x 2 testuser testuser ... target/debug/agent

# After watchdog install
-rwxr-xr-x 2 root root ... target/debug/agent
-rwxr-xr-x 1 root root ... /etc/init.d/synthaea-agent
```

**Verified on Alpine VM:**
- ✅ Non-root-owned binary normalized to root:root
- ✅ OpenRC script created as root:root
- ✅ cargo test -p watchdog: 7/7 passing
- ✅ clippy clean

---

## Platform Coverage ✓

**Linux-specific:**
```rust
#[cfg(target_os = "linux")]
pub(crate) fn harden_ownership(path: &Path) -> std::io::Result<()>
```

Correctly gated - this is Linux-specific behavior.

**macOS tracked:**
> "macOS Endpoint Security path — still open, tracked in #103"

Appropriate - different security model on macOS.

---

## Consistency ✓

**Applied everywhere permissions are hardened:**
1. ✅ Agent binary (`resolve_paths`)
2. ✅ Watchdog binary (`resolve_paths`)
3. ✅ Systemd unit (`install_systemd`)
4. ✅ OpenRC script (`install_openrc`)

No gaps - every `harden_permissions` call now has a corresponding `harden_ownership`.

---

### 1. Ordering
Current: `harden_permissions` then `harden_ownership`

This is fine, but worth noting: if `harden_ownership` fails, the permissions are already hardened (no rollback). This is acceptable because:
- Both operations are idempotent
- Partial hardening is better than none
- Proper error propagation means install fails loudly

### 2. Test coverage
The unit test only verifies `uid()` and `gid()` are 0. Could also verify they were *changed* from non-zero:

```rust
// Optional enhancement (not blocking):
std::os::unix::fs::chown(&path, Some(1000), Some(1000)).unwrap();  // Make non-root
harden_ownership(&path).unwrap();  // Should change to root
assert_eq!(meta.uid(), 0);
```

But current test is sufficient - it proves the function works.

---

## Verdict

This is a clean, well-scoped security fix that closes a real gap. The implementation is minimal, correct, and well-tested.

**Impact:**
- ✅ Closes ownership bypass in tamper protection
- ✅ Completes #103 hardening for Linux
- ✅ No breaking changes
- ✅ No new dependencies

**Recommendation:** Merge immediately.

**Follow-up:** #103 remains open for macOS Endpoint Security path (as noted in PR).

PR #271, opened 2026-09-21
Wire transport (run --server) + the architecture backbone docs

florianamette:
Two pieces, one theme: give the agent its composition spine and write down the architecture it implements.

## Transport wired (#24 foundation)

`run --server <url>` turns on store-and-forward upload: events append to a byte-capped spool next to the alerts file (on the enrich worker — capture never does file I/O), and a dedicated thread drains it to `POST /api/v1/ingest/events` with backoff. Without the flag, behavior is byte-identical to before.

Wiring the orphan crate immediately surfaced two interface bugs that made its promised at-least-once delivery unimplementable — the argument for wiring things: `EventDrain` had no `ack` (so a drain could only be at-most-once or never-complete), and `UploadLoop::stop(&mut self)` was uncallable while `run(&mut self)` blocked. Both fixed: the drain contract is now drain/ack/skip (skip = poison-segment escape on non-retryable server rejection), the loop stops via the sensors' shared-`AtomicBool` idiom. The health beacon now reports real spool depth/shedding and POSTs to the heartbeat endpoint (Linux, best-effort).

`commands/common.rs` is the new platform-independent `run` spine — the answer to "every new pipeline stage lands twice": transport, sink, banner, and progress heartbeat compose once; the platform modules keep only sensor selection and platform-only wiring.

Behavior pinned by tests: `unacked_drain_redelivers_acked_drain_deletes`, `skipped_segment_is_discarded_and_newer_data_flows`, `empty_drain_is_never_acked`.

## Architecture backbone (docs/architecture/)

`overview.md` and `agent.md` go from stubs to the real backbone: system layering, the three traffic planes with honest live-vs-planned markers, kill-resistance layers, the eight invariants future changes must not break, the agent's threading model, degraded modes, and on-disk layout.

## Verification

Full gauntlet clean (fmt, deps, clippy host + linux target + windows sensor crates, 83 test targets, deny, docs). The agent's own msvc cross-check now hits the known `ring` build-script limitation from macOS (transport is a new agent dep) — CI's native windows job covers it; the `windows.rs` changes are minimal (signature + pipeline call).


Jihair54 (commented):
Reviewed the diff — solid, well-tested design (the at-least-once drain/ack/skip contract, the shared-AtomicBool stop idiom, the honest degraded-mode docs). Two non-blocking observations:

**1. `EventDrain::skip()` discards the whole drained segment, not just the poison sub-batch.**

`upload_once` chunks the drained `Vec<Event>` into `batch_size` pieces (`crates/transport/src/upload.rs:102`). If chunk 2 of 5 gets a permanent rejection, `self.drain.skip()` discards the *entire* segment `drain()` returned — including chunks 3-5, which were never attempted and might have been valid. Probably fine in practice if segments are usually ~one batch, but worth confirming that's the expectation (or documenting it) since it's a silent-loss path for otherwise-good data sharing a segment with a poison record.

**2. `UploadLoop::stop_handle()` doesn't appear to be wired to shutdown anywhere.**

It fixes a real bug (`stop(&mut self)` was uncallable while `run(&mut self)` blocked), but I don't see it called from `agent/src/upload.rs::start` or either platform's `cmd_run` — no Ctrl-C handler signals it. Not a correctness issue (at-least-once means nothing is lost, the thread just dies with the process instead of stopping gracefully), but the fix seems to stop short of being exercised outside its own unit tests. Might be worth wiring on a follow-up if a graceful final-flush-before-exit matters.

Nothing here blocks merge from my read — flagging for awareness / possible follow-up.

old-dov:
@florianamette Heads up — this PR is now showing as conflicting with `main`. #287 (just merged) also touched `agent/src/commands/linux.rs`, which is likely the source. Needs a rebase before it can merge.

old-dov:
Reviewed and pushed a small fix (7222441): `RunPipeline::transport` failed `cargo clippy -p agent --all-targets -- -D warnings` on a native Windows build (`windows.rs` only reads `pipeline.sink`, not `pipeline.transport` — field was dead on that target). Added the same `#[cfg_attr(not(target_os = "linux"), allow(dead_code))]` already used on `TransportHandle::client`. Also committed the `Cargo.lock` update for the new `store`/`transport` deps in `agent/Cargo.toml`, which was missing.

Otherwise validated: clippy clean (Linux via WSL + native Windows) and `cargo test -p agent -p transport --all-targets` green (26/26) on Linux.

**Blocking issue for merge:** this branch now has a real conflict with `main`, not a mechanical one. `cmd_run`'s signature grew independently in two directions since this PR's base:
- here: `server: Option<&str>`
- on `main` (looks like PR #272, sensor-linux-uprobes wiring): `enable_tls_capture: bool, enable_readline_capture: bool`

Both need to land together in `agent/src/commands/{mod,linux,windows}.rs` and `main.rs`'s CLI wiring, in whatever order/shape keeps both features correctly threaded through `cmd_run`. I didn't want to guess at the intended composition of two features I didn't write, so leaving this for you to resolve rather than merging a guess. Happy to help once you've picked the shape.

florianamette:
Resolved — thanks both.

**Composition** (@old-dov): merged `main` into the branch (no rebase, you'd pushed to it). Both features land with each in its architectural home: `enable_tls_capture`/`enable_readline_capture` are Linux-sensor configuration, so they stay in the platform arm and thread into `spawn_uprobes_sensor` exactly as #272 wired them (inert on Windows, same pattern as the response flags); `--server` is pipeline-spine configuration and flows through `commands::common::wire_run_pipeline`. `cmd_run` carries both: `(alerts, events, enable_kill, enable_quarantine, enable_tls_capture, enable_readline_capture, server)`. Thanks for 7222441 — kept as-is.

**skip() granularity** (@Jihair54, obs 1): correct read, and it's now documented on `EventDrain::skip` rather than left implicit — the discard unit is the drained segment, sub-segment retry isn't expressible in the spool's two-phase protocol, and one segment (1 MiB cap ≈ one upload batch) is the accepted blast radius for the rare permanent-rejection path.

**stop_handle wiring** (obs 2): agreed it stops short — graceful final-flush-on-shutdown needs the agent to gain a coordinated shutdown path first (the health/upload threads all currently die with the process, per the note in `upload.rs`). Follow-up material; the handle exists so that wiring doesn't need another transport change.

Full gauntlet green on the merged result (fmt, deps, clippy host+linux+windows-sensor targets, tests, deny, docs).

PR #273, opened 2026-09-21
feat(sensor-linux-audit): classify SELinux AVC denials

old-dov:
Follow-up to the Linux telemetry matrix (#268), which flagged this as a near-free win.

## What

The audit fallback sensor (#247) already subscribes to the whole `NETLINK_AUDIT` multicast stream for execve/connect — SELinux AVC records (`type=AVC`, 1400) were already arriving at the socket, `classify()`'s `_ => None` arm just dropped them. Adds a match arm and `AuditEvent::PolicyDenial { comm, scontext, tcontext, tclass, permissive }`.

## The pid gap, found by tracing the parser by hand

No `pid` field on this variant. The kernel's AVC message body is `avc:  denied  { read } for  pid=1234 comm="httpd" ...` — text, not `key=value`, before the first real field. The shared byte-scanning field parser (`parse_fields`) has no concept of this preamble, so it folds `"avc:  denied  { read } for  pid"` into a single corrupted key, and only the *value* `1234` survives as unattributed. I traced this by hand against a real-format AVC line: every field *after* that first one (`comm`, `scontext`, `tcontext`, `tclass`, `permissive`) parses with a clean key — the corruption is scoped to exactly the first token. Not worth a preamble-skipping special case in the shared parser (used by the already-validated EXECVE/SOCKADDR paths) for one supplementary record type.

Always classified as a denial: the "denied"/"granted" word lives inside that same corrupted first field, and `auditallow`-logged grants are rare in practice — not worth recovering for that.

## Not wired to schema::Event

Same posture as the journal sensor's unit-lifecycle events (issue #93, and now documented the same way there): a Linux-only "policy denial" shape would preempt a cross-platform decision nobody's made (Windows AppLocker/WDAC and macOS TCC/Gatekeeper denials are the same concept). Traced via `tracing::info!` instead of dropped silently — that's the actual improvement here, not a full pipeline wire-up.

## Also

Fixed this crate's own doc comment, which claimed fanotify as a source — it isn't and never was (see #268's matrix).

## Validation

WSL: `cargo test -p sensor-linux-audit` (18/18, including 3 new tests against a real AVC line's field shape) and `cargo test -p sensor-linux-audit -p agent` (agent's own 19 tests + the LSM io_uring integration test still pass, confirming the `sensor.rs` dispatch change doesn't regress anything). `cargo clippy -p sensor-linux-audit --all-targets -- -D warnings` clean.

Not yet validated against a *real* AVC record from an actual SELinux-enforcing host (no such lab box exercises live probe loading today — see #268's "server-distro gap" section) — the field-shape reasoning above is traced by hand against the documented kernel format, same discipline as #247's own pre-real-hardware review, not yet a "confirmed on real hardware" claim.


Jihair54 (approved):
Clean, well-scoped PR — reviewed the diff carefully.

- The `match event` → `match &event` change is correct: needed to bind `PolicyDenial`'s fields without moving `event`, and the existing `Exec`/`Connect` arms still work unchanged against the reference.
- `Option<Event>` return threaded correctly through to `sink.on_event` — no silent drop, no panic on the new arm.
- The pid-field corruption reasoning (preamble text folded into the first key) is traced clearly in the doc comment and matches how the shared `parse_fields` byte scanner works; the `scontext`/`tclass`-required guard in `classify_avc` is a sensible floor against an all-`None` event.
- Three new tests cover the real shape (corrupted first key, clean rest), permissive mode, and the drop-when-empty case.
- Honest about what's not yet true (no real SELinux-enforcing host validated this end-to-end) rather than overclaiming — appreciate that.

Nothing to block on. Approving.

PR #287, opened 2026-09-21
feat(sensors/linux-journal, rules, schema): map systemd unit first-start to persistence (issue #93)

old-dov:
## Summary

Closes the second half of issue #93 ("A service install/start is
case-attachable context") and resolves the cross-platform decision
`sensor-linux-journal::auth`'s module doc had deliberately left open: now
that issue #224 (Windows' own service-lifecycle event reconciliation)
landed today, `sensor-windows-eventlog`'s 7045 handling confirms the
precedent to follow — reuse `FileOpenEvent` + a flag bit for persistence
signals, not a new `schema::Event` variant.

- `schema::FLAG_PERSISTENCE_SYSTEMD_ARTIFACT` (new bit, not schema-version-visible — same footing as the three existing `FLAG_PERSISTENCE_*` constants)
- `sensor-linux-journal::persistence::UnitPersistenceTracker` — maps a systemd unit's first-observed start into `FileOpenEvent`
- `rules::check_systemd_service_persistence` — T1543.002 (Create or Modify System Process: Systemd Service), the Linux sibling of the existing T1543.003 rule
- Wired into `agent::commands::linux::spawn_journal_tail` alongside the existing `to_auth_event` mapping

## Why a tracker, not a direct 1:1 mapping

journald's `JOB_TYPE=start`/`JOB_RESULT=done` fires on **every** start of a
unit — install or routine restart alike. Windows' 7045 is different: the
Service Control Manager only ever writes it once, at actual registration.
Mapping every `UnitStarted` straight to a persistence alert would flag every
`cron.service` restart as a new artifact — a correctness bug, not fidelity
to the Windows shape.

`UnitPersistenceTracker` closes that gap pragmatically: it keeps a
per-agent-process set of unit names already seen, and only reports the
*first* start of each. This is a documented approximation of "just
installed," not a true signal — a unit already running before the agent
started still alerts once, and an agent restart forgets what it had already
seen (same known gap as this crate's cursor-persistence limitation, see the
crate doc). Good enough for issue #93's bar; a real install-vs-restart
signal would need more (e.g. correlating with the unit file's mtime).

## Test plan

- [x] `cargo test -p schema -p rules -p sensor-linux-journal -p agent` — all green (WSL)
- [x] `cargo clippy -p schema -p rules -p sensor-linux-journal --lib --tests -- -D warnings` and `-p agent --bins --tests` — clean
- [x] `cargo test -p correlator -p ml` — no regression from the reused `flags` bit
- [x] `cargo test -p schema` (golden fixtures + `v1_compat`) — confirms no `SCHEMA_VERSION` bump needed, as expected for reusing an existing `u32` field
- [ ] Review


Jihair54 (approved):
Reviewed the diff carefully.

- The design decision (reuse `FileOpenEvent` + a distinct `flags` bit, no new `Event` variant) correctly follows the precedent #224 just established for the Windows side — good cross-platform consistency, and the module docs on both `auth.rs` and the new `persistence.rs` explain the reasoning clearly rather than just asserting it.
- `FLAG_PERSISTENCE_SYSTEMD_ARTIFACT = 0x4000_0000` is a genuinely distinct bit from the three existing `FLAG_PERSISTENCE_*` constants and from every real `open(2)` flag (those all stay well under `0x00500000`) — no collision risk.
- `UnitPersistenceTracker::observe`'s dedup logic is correct: `HashSet::insert` returning `false` (already present) short-circuits to `None`, matching `second_start_of_the_same_unit_is_not_reported_again`. The `unit.clone()?` early-return on `None` is clean.
- Honest about the approximation: journald fires on every unit start (install or restart alike), so "first start this agent process has seen" is explicitly documented as not a true install signal, with the agent-restart-forgets-what-it-saw gap called out in three places (flag doc, module doc, crate doc) rather than glossed over.
- Test coverage is thorough: start/stop/failed discrimination, missing unit name, independent per-unit tracking, and the defensive "both a systemd and a Windows bit set" case, even though the PR itself notes that case can't happen in practice (two sensors never share a process).

Nothing to block on. Approving.

Issue #23, opened 2026-09-03
Build policy: shared policy model


florianamette:
**Component:** `crates/policy` · **Source:** new development

Versioned, signed policy: active rules/models, permitted response actions, thresholds, per-host overrides. Base-tier crate (everything may depend on it). Signature verification on the agent; the server signs.

## Done when
- [ ] Policy types serialize/verify round-trip with test vectors
- [ ] Response actions consult policy (deny-by-default)

PR #299, opened 2026-09-22
feat(policy): shared policy document — schema, canonical JSON, Ed25519, layered overrides (#23)

Sollykhan:
## Summary

Delivers the `policy` crate work of #23 on the basis of ADR-0010 (shared policy model) and ADR-0011 (override granularity). The four pre-existing masquerade helpers (`is_trusted_system_path`, `expected_parents`, `parent_exclusion_applies`, `name_exclusion_applies`) and the two flat structs (`EventLogPolicy`, `ResponsePolicy`) are **kept unchanged** — v1 coexistence per ADR-0010 §Consequences (stepping stone).

## New modules

| Module | Purpose | LOC + tests |
|---|---|---|
| `document` | `Policy = PolicyMetadata + PolicyPayload` — the on-the-wire shape; v1 sections: `sensors` (map-typed, holds `windows_eventlog` with safety-critical `redaction` sub-doc), `response` (flat), `compliance_mode` (flat, safety-critical), `experimental` (open map); empty placeholders for `rules` / `models` / `thresholds` | ~360 |
| `canonical` | `to_canonical_bytes` — deterministic JSON (indent=2, sorted keys, `signature` → `""`). **The shared canonicalizer both ADR-0010 §3 and ADR-0015 §Decision 2 call for** — the updater manifest (issue #30) will reuse this same function | ~200 |
| `signature` | `sign` + `verify` via `ring` Ed25519. Signature format = **lowercase-hex 128 chars** (aligned with SHA-256 encoding in `tamper::integrity`). `SIGNATURE_LEN_BYTES = 64`, `PUBLIC_KEY_LEN_BYTES = 32` | ~250 |
| `merge` | `apply_overrides` — the three granularities from ADR-0011: flat / map-typed / safety-critical. Metadata is taken entirely from the override (to preserve the anti-rollback mechanism via `policy_version`) | ~250 |
| `error` | `PolicyError` with 7 distinct variants — every failure case is named and actionable | ~110 |

### Safety-critical: schema-level, not a runtime walker

ADR-0011 §Decision 2 requires that a safety-critical sub-object present in an override be **complete**, otherwise rejection at parse time. The naive implementation is a runtime walker over `SAFETY_CRITICAL_PATHS`.

**This PR takes a different route**: every schema-declared field of a safety-critical sub-object is marked serde-required (no `#[serde(default)]`). A partial override therefore fails to deserialize as `RedactionPolicy` at all — it surfaces as `PolicyError::Parse` with the missing-field name serde itself produced. That is:

- Stricter than a hand-rolled walker (the error names the exact missing field, without a duplicated list to keep in sync)
- Earlier in the load path (before merge)
- Zero cost in the happy path

`SAFETY_CRITICAL_PATHS` stays as a **declarative constant** — it documents intent, matches the reference form ADR-0011 §Decision 3 called for, and is guarded by a regression test (`partial_redaction_override_fails_at_parse_not_at_merge`) that fails if a future contributor adds `#[serde(default)]` to a safety-critical field.

### ADR ambiguities settled

- **Signature encoding**: `lowercase-hex 128 chars`. ADR-0010 §3 does not pin an encoding; aligned with SHA-256 encoding in `tamper::integrity::Manifest`. Uppercase is explicitly rejected to keep the format contract single-valued.
- **`issued_at` unit**: nanoseconds `u64` since Unix epoch. Same unit as `schema::EventMeta` timestamps everywhere in the workspace; no new `chrono`/`time` dependency.
- **`ring` vs `ed25519-dalek`**: `ring 0.17` — already a transitive dep via `rustls` (confirmed in Cargo.lock). Same dispatch as ADR-0015 §Decision 3.

### Minimal v1 payload scope (documented in `document.rs`)

The four sections `sensors` / `response` / `compliance_mode` / `experimental` each exercise a distinct merge granularity. `rules`, `models`, `thresholds` are empty placeholder structs — they land with a consumer, additive non-breaking.

## Test plan

- [x] `cargo test -p policy` — **38/38 tests OK** (12 pre-existing + 26 new) in 27s cold build, 0.02s test run.
- [x] `cargo clippy -p policy --all-targets -- -D warnings` — clean.
- [x] `sign_then_verify_succeeds` round-trip using a test keypair generated on the fly via `ring`.
- [x] `signature_survives_serde_round_trip` — the real production shape (control plane signs and serializes; agent parses and verifies).
- [x] `tamper_in_payload_breaks_verification` — flipping `policy_version` after signing invalidates the signature.
- [x] `partial_redaction_override_fails_at_parse_not_at_merge` — safety-critical guard.

## Follow-ups (deliberately out of scope)

- **Wire `PolicyPayload.sensors.windows_eventlog` → `EventLogConfig`** in `agent/src/commands/windows.rs` (replacing the `EventLogPolicy` glue) — ADR-0010 §Consequences flags this as the implementation follow-up.
- **Key custody / rotation** — ADR-0010 §Deferred, lands with the M8 control plane.
- **Real fields in `RulesSection` / `ModelsSection` / `ThresholdsSection`** — land with a consumer.

Refs #23. ADRs: 0010, 0011.


old-dov:
Reviewed the crypto/canonicalization side (signature.rs, canonical.rs) closely — solid, well-tested, no issues (signature blanking, key-sort determinism, timing-neutral verify all check out).

**Found and pushed a fix (27b3fe5) for a real gap in the safety-critical mechanism**, which is this PR's core selling point, so wanted to flag it clearly rather than let it slide through.

## The gap

ADR-0011's motivating example is explicit: *"an operator overriding `sensors.windows_eventlog` and forgetting `redaction.pii_scrub_enabled`, silently turning PII scrub off — the exact regression Hugo flagged... on #178."*

But `redaction: Option<RedactionPolicy>` is `#[serde(default)]` (`document.rs:210`), so an override that omits the field **entirely** parses fine — the parse-time safety-critical check only catches a *present-but-incomplete* `redaction` (e.g. `{}`), not an *absent* one. And since `apply_sensors` replaced `windows_eventlog` as a whole object (not field-by-field), any override that only meant to touch e.g. `service_installs_enabled` silently reset `redaction` to `None` on merge — reproducing the exact #178 scenario the ADR exists to close, just one layer later (survives merge even though a *partial* redaction object can no longer survive parse).

No existing test caught this — `map_typed_section_only_touches_named_key` always includes `redaction` explicitly in its override, so it never exercises the omission path.

(Smaller/related: `PolicyError::SafetyCriticalPartial` is a fully-written error variant that's never actually constructed anywhere in the crate — looks like a leftover from the "runtime walker" approach the module doc says was deliberately rejected in favor of serde-required fields. Didn't touch it since it's not load-bearing, but worth a cleanup pass at some point.)

## The fix

`apply_sensors` now inherits baseline's `redaction` when the override's `windows_eventlog` doesn't mention it, and only lets an override change `redaction` by supplying a complete `RedactionPolicy` (already guaranteed complete by the existing parse-time check). An operator who wants to explicitly disable it still can — `{"pii_scrub_enabled": false}` is a complete object, unambiguously distinct from omitting the field.

Added two regression tests: the exact #178 scenario (override touches an unrelated toggle, omits redaction, baseline's redaction must survive), and the explicit-change counterpart (override *can* still turn redaction off/on when it says so). `cargo test -p policy` — 40/40 green (38 existing + 2 new). `cargo clippy -p policy --all-targets -- -D warnings` — clean.

Sollykhan:
Thanks for the catch, @old-dov — a real gap and I appreciate you pushing the fix rather than just flagging it. Reviewed the patch, the reasoning holds up on both fronts:

1. **The trap is real.** The parse-time carve-out for `RedactionPolicy` (serde-required fields, rejects `{}` at parse) only protects against a *present-but-partial* `redaction`. The whole-object replace in `apply_sensors` sliding an *omitted* field to `None` at merge time silently reopens exactly the #178 window. I missed that a `#[serde(default)] Option<RedactionPolicy>` makes "absent" and "explicitly `None`" indistinguishable after parse — so the invariant I thought was enforced by "safety-critical completeness" was really only enforced when the field was mentioned at all.

2. **The fix threads the needle.** Inheriting baseline's `redaction` when the override omits it, while keeping the parse-time check for any *supplied* `RedactionPolicy`, preserves both properties: the #178 scenario cannot silently drop scrub, and an operator who genuinely wants to turn it off still can via `"redaction": {"pii_scrub_enabled": false}` — a complete object, semantically distinct from omission. The two new tests cover both branches, including the exact #178 scenario I hadn't parametrized.

Rebased on main (Cargo.toml conflict resolved — `tokio` from #308 and `ring` from this PR coexist), 40/40 passing.

## Small follow-up (not scope for this PR)

`WindowsEventlogSensorPolicy`'s other fields (`service_installs_enabled`, `scheduled_tasks_enabled`, `account_creations_enabled`, `logon_events_enabled`) are `bool`, not `Option<bool>`, so their exposure to the same "override omits field" trap depends on whatever `#[serde(default)]` attrs sit on the struct. Worth a pass with the same lens (present-but-incomplete vs. absent) even for non-safety-critical fields — the failure mode is the same shape: override an unrelated toggle, silently reset an intentionally-off detection to its default-on (or vice versa). Not blocking this PR; capturing it here so it doesn't get lost.

Issue #326, opened 2026-09-22
sensor-linux: high-frequency write(2) activity appears to starve other file-event ring buffers


Jihair54:
**Component:** \`crates/sensors/linux/userspace\` · **Priority:** 🟠 High
**Discovered while validating:** #262 path-filtering (PR #325)

## What was observed

Live-capturing events on the Alpine lab VM (\`agent capture-events\`) while Docker was running: a single window captured **5,849,407 \`file_write\` events and exactly 1 \`file_open\` event**, despite deliberately performing several real \`open\`/read operations during the capture (a plain \`echo >\`/\`cat\` on a path outside every filtered prefix). The write volume came from Alpine's OpenRC \`log_proxy\` helper for \`dockerd\` (\`log_proxy -m /var/log/docker.log\`), writing a tiny control file (\`/var/log/.docker.log.control\`) in an extremely tight loop — ordinary system infrastructure, not test noise or anything malicious.

Stopping the \`docker\` service (removing \`log_proxy\`) was tried to get a clean read, but the same symptom (overwhelming event volume, terminal flooded, a live \`capture-events\` run left running 2+ minutes with zero bytes flushed to its output file) reappeared — **so \`log_proxy\` may not be the only source, or there's a second contributing factor not yet identified.** Root cause is not confirmed; this issue is filed to track investigating it, not to assert a specific fix.

## Why this matters beyond the raw ring-buffer-full warning

\`sensor-linux\`'s \`FILE_WRITE_EVENTS\` ring buffer filling up and dropping events under load is expected and already handled (\`RingBuf::output()\` returns \`Err\`, logged as a \`warn!\` and dropped — see \`crates/sensors/linux/ebpf/src/main.rs\`'s \`try_sys_enter_write\`). What's more concerning is that **other, independent ring buffers** (\`FILE_OPEN_EVENTS\`, \`FILE_DELETE_EVENTS\`, \`FILE_RENAME_EVENTS\`, \`FILE_CHMOD_EVENTS\`, \`FILE_CHOWN_EVENTS\`, ...) appeared to go almost entirely undrained in the same window, even though they are backed by separate eBPF maps and should be unaffected by \`FILE_WRITE_EVENTS\` specifically filling up.

The userspace drain loop (\`crates/sensors/linux/userspace/src/sensor.rs\`, \`LinuxSensor::run_async\`) is a single \`tokio::select!\` across one \`AsyncFd::readable_mut()\` future per ring buffer, all in one task:

[...]

PR #331, opened 2026-09-22
fix(sensors/linux): bound ring-buffer drain to prevent event-type starvation (issue #326)

old-dov:
## Root cause

While reviewing #326 (5,849,407 `file_write` events vs. exactly 1 `file_open` in the same capture window, discovered validating #325), I traced it to source: `aya::maps::RingBuf::next()` is a synchronous read of live shared mmap'd memory — no syscall, no `.await`, it just re-checks the producer position against the consumer position every call.

Both drain sites' inner loops have no yield point:

```rust
while let Some(item) = rb.next() {
    // ... decode + emit, no .await anywhere in here
}
guard.clear_ready();
```

Under a sustained producer (a tight `write(2)` loop — Alpine's `log_proxy` for `dockerd` writing a control file, per #325's investigation), `rb.next()` can keep returning `Some` indefinitely. Since neither `rb.next()` nor the loop body ever awaits, **the whole task never returns control to `tokio::select!`**, so the other 9 branches in `sensor-linux`'s loop (2 in `sensor-linux-uprobes`) — and the reactor's own epoll processing for their fds — never get polled. This is outright starvation, not the "statistical disadvantage under extreme skew" #326 speculated might be the mechanism; the observed ratio (millions to one) is consistent with a task that simply never comes back to check the other branches for the entire capture window.

## Fix

Cap each `drain!`/`drain_tls!`/`drain_readline!` call at `MAX_ITEMS_PER_DRAIN` (256) items, then return to `select!` regardless of whether the ring buffer is empty yet:

```rust
let mut drained = 0usize;
while drained < MAX_ITEMS_PER_DRAIN {
    let Some(item) = rb.next() else { break };
    drained += 1;
    // ... decode + emit, unchanged
}
if drained < MAX_ITEMS_PER_DRAIN {
    guard.clear_ready();
}
```

Only clearing readiness when the buffer actually ran dry (`drained < cap`) matters: if we stopped because we hit the cap, there's still data waiting, so the ready-flag stays set and the next `select!` iteration re-polls this branch immediately (continuing the drain) rather than blocking for a fresh epoll edge that might not come for a while. Critically, returning to the top of the `loop` re-issues `.readable_mut()` on *every* branch each time — a real tokio I/O primitive that participates in tokio's cooperative-scheduling budget, so a task that keeps "instantly" resolving its poll eventually gets forced to yield, letting the reactor actually process the other fds' epoll events. The unbounded version never gave that budget mechanism a chance to engage at all.

Applied to the three sites with this exact pattern:
- `crates/sensors/linux/userspace/src/sensor.rs` — `drain!` (all 11 ring buffers: exec, file open/write/delete/rename/chmod/chown, connect, socket bind/listen/accept)
- `crates/sensors/linux/uprobes/src/sensor.rs` — `drain_tls!` and `drain_readline!`

## Validation

- `cargo test -p sensor-linux -p sensor-linux-uprobes --all-targets` — 56/56 + 47/47, unchanged, no regressions.
- `cargo clippy -p sensor-linux -p sensor-linux-uprobes --all-targets -- -D warnings` — clean except one pre-existing, unrelated `must_use_candidate` on `normalize::udp_send` that's already on `main` before this change (confirmed via `git stash` + re-run).
- `python3 tools/check-deps.py` — unaffected, no dependency changes.

**Not validated**: a live before/after repro of the starvation itself (synthetic tight `write(2)` loop + event-count comparison on the lab VM). No lab VM was available this session. The existing test suite has no ring-buffer-contention coverage to exercise this either way. This fix is grounded in `RingBuf::next()`'s documented synchronous semantics (`aya-0.14.0/src/maps/ring_buf.rs`) and tokio's documented cooperative-scheduling budget for `AsyncFd`, not in reproducing #326's failure and confirming it's gone. Flagging this gap explicitly rather than claiming a live-validated fix — whoever picks this up next on the lab VM, the repro described in #326 (tight write loop + `file_open`/etc. event counts) is the right test.

## Scope

Read-only fix to two existing files, no schema/wire changes, no new dependencies. `MAX_ITEMS_PER_DRAIN = 256` is a first guess, not tuned against real throughput numbers — worth revisiting once the live repro above is run.

Refs #326.


old-dov:
## Live validation on the lab VM (Rocky9, kernel 5.14.0-362.13.1.el9_3, x86_64)

Reproduced #326 directly and confirmed the fix. Built both `main` and this branch, ran the identical scenario against each: a ~1kHz `write(2)` loop on an already-open fd (mirroring `log_proxy`'s shape) running the whole time, plus 15 real file operations (`open`+`write`, `chmod`, `rename`, `delete`) spaced one per second, captured via `agent capture-events` for 20s.

| Event type | `main` (before) | this branch (after) |
|---|---|---|
| `file_write` | 12,113,301 | 13,168,709 |
| `file_open` | **0** | **315,536** |
| `exec` | 13 | 10,426 |
| `file_rename` | 0 | 16 |
| `file_delete` | 0 | 15 |
| `file_chmod` | 0 | 15 |

Before: total starvation — zero `file_open`/`rename`/`delete`/`chmod` events despite the script explicitly performing all four, and only 13 of the many real `exec`s (chmod/mv/rm/sleep subprocesses) got through. Confirms this was outright starvation, not statistical disadvantage — matches the code-level analysis in the PR description exactly.

After: every event type flows through at a plausible volume (the `file_open` jump to 315k reflects the VM's normal background activity finally being captured, not just my 15 test opens), `rename`/`delete`/`chmod` land within 1 of the 15 real operations performed, and `file_write` throughput is unaffected (if anything slightly higher — the fix costs nothing on the dominant path).

This closes the "not validated" gap noted in the original PR description — happy to merge whenever a second pair of eyes has looked at the diff.

Jihair54:
## Live repro validated

Filled the one gap flagged in this PR's description ("Not validated: a live before/after repro of the starvation itself") on the Alpine lab VM.

**Setup:** same flood source as #326's original discovery (Docker's `log_proxy` writing its OpenRC control file in a tight loop). Built this branch, ran `agent capture-events` (raw `JsonlEventSink`, no rule filtering) under `sudo timeout 25`, and during the window generated a known set of file events outside the filtered paths (`echo >`, `cat`, `chmod 600`, `mv`, `rm` × 3, on `/root/edr-new-test-*`, not `/tmp`).

**Result — 25s window, 1,551,599 total events:**

| type | count |
|---|---|
| `file_write` | 1,551,464 |
| `file_open` | 109 |
| `exec` | 16 |
| `file_chmod` | 3 |
| `file_rename` | 3 |
| `file_delete` | 3 |
| `connect` | 1 |

All 3 synthetic chmod/rename/delete events came through (3/3 each), plus 109 organic `file_open` and 16 `exec` events from normal system activity — despite `file_write` still being 99.99% of the volume. Before this fix, the same flood source (documented in #326) produced 5.8M `file_write` against exactly **1** `file_open` in a comparable window — the other event types were effectively invisible. The 256-item drain cap clearly restores fairness across event types even under sustained one-sided producer pressure.

Also confirms the capture itself terminates cleanly under `timeout` with no hang — the earlier (pre-fix) investigation needed a second SSH session and `pkill -9` to recover from an unresponsive shell under the same load.

Didn't get to tune `MAX_ITEMS_PER_DRAIN` against real throughput numbers, just confirmed the mechanism works as designed. LGTM from a behavioral standpoint.

Jihair54 (approved):
## Code review

Reviewed the diff across all three modified sites (`drain!` in `sensor-linux`, `drain_tls!`/`drain_readline!` in `sensor-linux-uprobes`).

**Logic:** `while drained < MAX_ITEMS_PER_DRAIN { let Some(item) = rb.next() else { break }; drained += 1; ... }`, then `clear_ready()` only when `drained < MAX_ITEMS_PER_DRAIN` — correct. If the loop exits because the buffer ran dry, clearing readiness is right (nothing left to poll). If it exits because the cap was hit, leaving the ready-flag set means `select!`'s next iteration re-polls this branch immediately rather than waiting for a fresh epoll edge that might not come — that's exactly what lets the drain continue across iterations without starving the other branches in between. Edge case (cap hit exactly as the buffer empties) just costs one harmless extra immediate re-poll that clears cleanly; no correctness issue.

Consistent across all three drain sites, `unsafe`/`SAFETY` blocks untouched, no behavior change to decoding or emission — this is a pure control-flow bound. The `MAX_ITEMS_PER_DRAIN` const is duplicated between the two crates rather than shared, but that's correctly forced by the dependency-direction rule (sensor crates depend only on `schema`), not an oversight.

Combined with the two independent live validations (mine and yours, on different VMs/kernels, both showing total-starvation-before vs. fair-flow-after with matching qualitative results) — this closes both the behavioral and code-correctness gaps. Approving.

PR #277, opened 2026-09-21
feat(sensors): Linux file write/delete/rename telemetry (#262)

Jihair54:
Phase 1 slice of #262: `write(2)`, `unlink(2)`/`unlinkat(2)`, and
`rename(2)`/`renameat(2)`/`renameat2(2)` telemetry on the Linux eBPF sensor,
mirroring the existing `sys_enter_open{,at}` pipeline end to end (wire →
eBPF tracepoint → userspace normalize → schema `Event` variant → golden
fixture). chmod/chown/xattr (the issue's Phase 2/3) and the perf/ransomware-
scenario validation are left as follow-up on the issue.

## Design note: `FileWrite` carries no path

`write(2)`/`pwrite64(2)` take a file descriptor, not a path, and this sensor
resolves no fd→path mapping (no `bpf_d_path`/LSM hook, no userspace
`/proc/<pid>/fd/<n>` lookup) — same spirit as `FileOpenEvent`'s existing
"not resolved against `dfd`" limitation. `FileWriteEvent` is therefore a
volume/frequency signal (pid/fd/bytes_requested) for burst-write detection
(ransomware, mass tampering), not a per-write path trail. `FileDeleteEvent`/
`FileRenameEvent` carry real paths straight off the syscall — no fd
resolution needed there, and the classic ransomware rename-to-`.locked`
signal lives entirely in `FileRenameEvent::new_path`.

Only plain `write(2)` is attached in this pass; `writev`/`pwrite64`/`pwritev`
are deferred (noted on the issue).

## Verification

Tracepoint offsets (x86_64/aarch64) are **verified**, not guessed: dumped
`/sys/kernel/tracing/events/syscalls/sys_enter_{write,unlink,unlinkat,rename,renameat,renameat2}/format`
on the Alpine lab VM (kernel 6.18.50-0-virt) before writing the eBPF code —
every offset matches the standard `syscalls:*` layout used elsewhere in this
file. i686 offsets are inferred by the same formula already used (and
flagged the same way) for `sys_enter_open`'s i686 table — not independently
verified on hardware.

`WIRE_VERSION` 5 → 6, `SCHEMA_VERSION` 14 → 15 (new `tests/fixtures/v15/`,
full copy-forward + 3 new fixtures). `sensor-linux-uprobes` has its own
separate `WIRE_VERSION` tripwire that also needed bumping — caught by a real
build failure on the lab VM, fixed in a follow-up commit on this branch.

[...]

Issue #262, opened 2026-09-21
sensor-linux: Add file write/delete/rename telemetry


SherlockOmss:
**Component:** `crates/sensors/linux/ebpf` · **Priority:** 🔴 Critical

## What

Currently `sensor-linux` only captures file **open** operations. Missing visibility into file modifications, deletions, and renames — core telemetry required for ransomware detection, data exfiltration, and tampering.

## Current State

**Tracepoints implemented:**
- ✅ `sys_enter_open`, `sys_enter_openat` (read/write opens)

**Missing:**
- ❌ File write/modify
- ❌ File delete
- ❌ File rename
- ❌ File permission changes (chmod/chown)
- ❌ Extended attributes (xattr)

### File Write (Critical)
- `sys_enter_write`
- `sys_enter_writev`
- `sys_enter_pwrite64`
- `sys_enter_pwritev`

### File Delete (Critical)
- `sys_enter_unlink`
- `sys_enter_unlinkat`

### File Rename (Critical)
- `sys_enter_rename`
- `sys_enter_renameat`
- `sys_enter_renameat2`

### Permission Changes (High)
- `sys_enter_chmod`
- `sys_enter_fchmod`
- `sys_enter_fchmodat`
- `sys_enter_chown`
- `sys_enter_fchown`
- `sys_enter_fchownat`

### Extended Attributes (Medium)
- `sys_enter_setxattr`
- `sys_enter_removexattr`

## Schema Changes

Add event types to `crates/schema`:
```rust
pub enum Event {
    // ... existing
    FileWrite(FileWriteEvent),   // NEW
    FileDelete(FileDeleteEvent), // NEW
    FileRename(FileRenameEvent), // NEW
    FileChmod(FileChmodEvent),   // NEW
    FileChown(FileChownEvent),   // NEW
}
```

## Detection Impact

**Enables:**
- Ransomware detection (rapid file modifications + renames to .encrypted)
- Data destruction (mass deletions)
- Log tampering (write/delete on log files)
- Privilege escalation (chmod +s on binaries)
- Anti-forensics (timestomping via touch/utimes)

**Examples:**
- Detect 100+ files modified in 10s window → likely ransomware
- Detect deletion of audit logs → tampering
- Detect chmod 4755 on unusual binary → SUID backdoor

## Performance Considerations

**Volume:** File write operations are HIGH frequency
**Mitigation:**
- Path filtering (skip /dev, /proc, /sys, temp dirs)
[...]

PR #332, opened 2026-09-22
feat(sensors/linux): xattr telemetry — setxattr/removexattr (issue #262 Phase 3)

Jihair54:
## What

Phase 3 of #262: adds `sys_enter_setxattr`/`sys_enter_removexattr` tracepoints, giving visibility into extended-attribute writes — relevant to capability-based privilege escalation (`security.capability`) and anti-forensics (`security.selinux`, `user.*` tampering).

## Changes

- **eBPF** (`crates/sensors/linux/ebpf`): `sys_enter_setxattr`/`sys_enter_removexattr` tracepoints, each capturing `path` + xattr `name` (max 64 bytes) from per-CPU scratch buffers. Path filtering (`is_filtered_path`, from #325) applied to both.
- **Wire** (`crates/sensors/linux/wire`): `FileSetxattrEvent`/`FileRemovexattrEvent`; `WIRE_VERSION` 11→12.
- **Userspace** (`crates/sensors/linux/userspace`): 2 new ring buffers + drain arms; `normalize::file_setxattr`/`file_removexattr`.
- **Schema** (`crates/schema`): `Event::FileSetxattr`/`FileRemovexattr`; `SCHEMA_VERSION` 19→20; `v20/` fixtures (copy-forward of `v19/` + 2 new golden fixtures); `fixtures::file_setxattr`/`file_removexattr` neutral builders.

## Deferred (out of scope for this PR)

- `fsetxattr`/`lsetxattr`/`fremovexattr`/`lremovexattr` variants (fd/symlink forms) — same rationale as the write-family deferrals in #277.
- xattr *value* capture — name-only, matching this issue's "Extended Attributes (Medium)" scope; value capture would need its own size/truncation design.

## Validation (Alpine lab VM)

- `cargo build -p sensor-linux -p agent` — clean
- `cargo test -p schema -p sensor-linux -p sensor-linux-uprobes` — 47/47 passed + doctests, including 2 new normalize unit tests (`security.capability`, `security.selinux`) and 2 new golden fixture tests
- `cargo clippy -p schema -p sensor-linux -p sensor-linux-uprobes -p agent --all-targets -- -D warnings` — clean (one unrelated pre-existing `must_use_candidate` finding on `udp_send`, fixed in the same branch)
- Tracepoint argument offsets (`pathname`@16, `name`@24, x86_64) verified against live `format` dumps on the lab VM before implementation

## Issue #262 checklist after this PR

- [x] Write/delete/rename tracepoints (Phase 1)
- [x] Schema events / userspace normalization
- [x] chmod/chown tracepoints (Phase 2)
- [x] Path filtering (#325)
- [x] xattr tracepoints (Phase 3) — this PR
- [ ] Performance validated on lab VM (<5% CPU overhead) — blocked on #326
- [ ] Ransomware scenario detection working end-to-end — not started


old-dov:
Heads-up: the starvation fix (#326) this checklist item is blocked on merged into `main` at 2046a66 (PR #331). Your branch was last synced before that — a rebase/merge of `main` should unblock the perf validation pass.

Jihair54:
Rebased onto \`main\` (\`7cd4b55\`) — \`MAX_ITEMS_PER_DRAIN\` (#331) is genuinely in this branch now, not just a clean-merge coincidence. Confirmed via grep + a fresh build/test/clippy/check-deps pass on the lab VM, all clean.

Also re-ran the #262 ransomware/CPU scenario from scratch on this rebased branch (the earlier numbers were measured before this rebase, so they technically predated #331's actual code — good catch). Results hold within noise: idle ~0.71% CPU (vs. 0.7% before), burst ~25.5% (vs. 16-29% range before), overall average ~2.3% (vs. 2.4% before). Posted as a follow-up on #262 too.

SherlockOmss:
crates/sensors/linux/wire/src/lib.rs:L1520: `MAX_XATTR_NAME_LEN = 64` vs kernel's `XATTR_NAME_MAX = 255`. Doc says "every name seen in practice is well under 64"—risky assumption. Silent truncation if exotic xattr appears (e.g. long `user.*` namespace). Consider: add truncation warn in eBPF probe or userspace normalize when `name_len` hits 64, so operator sees the edge case in logs rather than silent data loss.

Otherwise solid: path filtering applied, version bumps correct, golden fixtures comprehensive, normalize unit tests good.

Jihair54:
Good catch — fixed by capturing the kernel's actual XATTR_NAME_MAX (255) instead of the guessed 64-byte cap, rather than adding truncation detection on top of a cap that shouldn't need to exist. The wire struct is a fixed-size `#[repr(C)]` buffer either way (same pattern `MAX_PATH_LEN = 256` already uses in this file), so there's no real cost to just not truncating in the first place — and it removes exactly the silent-data-loss case you flagged, including the adversarial one (a long name chosen specifically to push the meaningful part past a small cap).

Pushed as a8ba355, validating on the VM now.

Issue #312, opened 2026-09-22
sensor-linux-uprobes: container attribution for TLS/readline events (residue of #80)


florianamette:
**Component:** `crates/sensors/linux/uprobes`

#80 delivered container attribution for the main sensor (cgroup-id capture + cgroupfs resolution + Docker socket lookup), but the uprobes sensor still passes `None`: `src/sensor.rs` carries two `TODO: container_id from /proc/<pid>/cgroup (issue #80)` markers pointing at the now-closed issue, and `src/normalize.rs` documents the id-only/image-name split as awaiting a follow-up that was never filed. TLS-plaintext and readline events are currently unattributable to workloads — exactly the event classes where container context matters most (a shell in a container vs on the host).

## Done when
- [ ] TLS/readline events carry `ContainerContext` with the same fidelity as the main sensor's events (shared `CgroupIdCache`/Docker-cache mechanism, now in `sensor-linux/src/container.rs`)
- [ ] The two stale `issue #80` TODOs are gone

old-dov:
Fixed by #335 (merged).

PR #335, opened 2026-09-22
fix(sensors/linux-uprobes): container attribution for TLS/readline events (issue #312)

old-dov:
## What

TLS-plaintext and readline events (`sensor-linux-uprobes`) had no `ContainerContext` — exactly the event classes where container context matters most (a shell in a container vs. on the host). `sensor.rs` carried two stale `TODO: container_id from /proc/<pid>/cgroup (issue #80)` markers pointing at a now-closed issue, and `normalize.rs` only ever produced an id-only context (image/name permanently `None`) — the follow-up #80's own doc comment said was awaited never got filed until now (#312).

Wires in the same mechanism the main `sensor-linux` crate uses: cgroup id → container id (cgroupfs walk, bounded LRU cache) → image/name (background Docker socket lookup, shared cache), at the same fidelity.

## Architecture decision (discussed before implementing)

`container.rs`/`docker.rs` are **duplicated** into `sensor-linux-uprobes`, not shared via a crate dependency. `sensor-*` crates may depend only on `schema` (+ their own platform wire crate) per `tools/check-deps.py` — `sensor-linux-uprobes` cannot depend on `sensor-linux`. The two realistic alternatives were:
- Move the ~500 lines of stateful LRU caching + Docker socket HTTP client into `sensor-linux-wire` — rejected, that crate's whole charter (`#![no_std]` ABI structs) is explicitly not a place for stateful/IO business logic.
- Add a new shared crate and edit `check-deps.py`'s dependency graph to allow it — a deliberate governance change affecting the whole team's understanding of the dependency rules, not something to do unilaterally inside an unrelated bugfix.

Duplication was the agreed call — same tradeoff already accepted for `MAX_ITEMS_PER_DRAIN` (#326), just at a larger scale here. Each file's module doc cross-references the other copy so a future change doesn't get made in only one place by accident.

## Changes

- **`container.rs`, `docker.rs`** (new): copied from `sensor-linux` verbatim, logic unchanged, with a module-doc note on the duplication and why.
- **`sensor.rs`**: `run_async` instantiates one `CgroupIdCache`/`DockerInfoCache` (matching `sensor-linux`'s pattern exactly), threads them into both `drain_tls!`/`drain_readline!` macros as new args, calls `container_context()` per event instead of hardcoding `None`. The two stale TODOs are gone.
- **`normalize.rs`**: `tls_capture`/`readline_input`/`meta` now take `Option<ContainerContext>` directly (the caller already resolved id + image + name) instead of `Option<String>` that could only ever produce an id-only context. Updated the two container-attribution tests to build/assert a full `ContainerContext` (id+image+name) instead of just an id.
- **`Cargo.toml`**: added `serde`/`serde_json` (`docker.rs`'s JSON parsing) and the `tokio` `io-util`/`time` features (`AsyncRead/WriteExt`, the lookup timeout) — the same set `sensor-linux` already carries, for the same reason.

## Validation

- `cargo build -p sensor-linux-uprobes -p agent` — clean.
- `cargo test -p sensor-linux-uprobes --all-targets` — **73/73** (up from 47 — the +26 are the copied `container.rs`/`docker.rs` test modules, unchanged from their `sensor-linux` originals).
- `cargo clippy -p sensor-linux-uprobes --all-targets -- -D warnings` — clean.
- `python3 tools/check-deps.py` — clean (38/38 crates).

No wire/schema changes — `EventMeta::cgroup_id` has existed since wire v4 (issue #204), this is purely userspace resolution logic that was never wired up on this sensor's side.

Refs #312.


Jihair54 (approved):

Read the diff (`sensor.rs`, `normalize.rs`, and skimmed `container.rs`/`docker.rs` against their claimed `sensor-linux` originals) and cross-checked the claims:

- `container_context(cgroup_id: u64, container_ids: &mut CgroupIdCache, docker_cache: &DockerInfoCache) -> Option<ContainerContext>` in the new `container.rs` matches the call sites in both `drain_tls!`/`drain_readline!` exactly (param types and order).
- `CgroupIdCache::new()` and `DockerInfoCache = Arc<Mutex<HashMap<...>>>` match how they're constructed in `UprobesSensor::run_async`.
- `normalize::meta`/`tls_capture`/`readline_input` signature change from `Option<String>` (id-only) to `Option<ContainerContext>` (full) is threaded consistently — no leftover call site still passing a bare id.
- The two stale `TODO: container_id from /proc/<pid>/cgroup (issue #80)` comments are genuinely gone, replaced with the real resolution call.
- Updated tests build a full `ContainerContext` (id+image+name) via the new `full_container_context()` helper and assert on the whole struct (`assert_eq!(e.meta.container, Some(ctx))`) instead of just the id — matches the "same fidelity as the main sensor" claim, not just a signature change with the same weak coverage.
- The `pub(crate)` visibility on `container.rs`'s items is correctly scoped (used only within this crate via `crate::container::{...}`), consistent with the dependency-direction rule cited in the PR body (`sensor-linux-uprobes` can't depend on `sensor-linux`, hence duplication rather than a shared crate) — read `tools/check-deps.py`'s `LEAF` set to confirm sensor crates are indeed restricted to `schema` only, so the "governance change" alternative really would have been out of scope here.

- `cargo build -p sensor-linux-uprobes -p agent` — clean
- `cargo test -p sensor-linux-uprobes --all-targets` — 73/73 passed (matches the PR's claimed count exactly)
- `cargo clippy -p sensor-linux-uprobes -p agent --all-targets -- -D warnings` — clean

No wire/schema changes to re-verify against golden fixtures, consistent with the "purely userspace resolution logic" claim. Approving.

Issue #313, opened 2026-09-22
sensor-linux-lsm: production wiring and remaining scope (residue of #91)


florianamette:
**Component:** `crates/sensors/linux/lsm`, `agent`

#91 closed with the observation-only foundation validated on real hardware, but the crate's own doc lists what was deliberately deferred, and none of it is tracked:

- **No production caller at all** — the crate is a dev-dependency of one agent integration test; `cmd_run` never loads it.
- **Inline blocking** (-EPERM on a verdict) — correctly blocked on #131/#133 (verdict model), noted here for the dependency.
- **`bprm_check`/`socket_connect` hooks** and a decided relationship between LSM-sourced and tracepoint-sourced file-open events (dedup question flagged in `crates/sensors/linux/ebpf/src/main.rs` near the `file_open` LSM program).
- **LSM-coverage capability reporting** — needs a schema surface decision (`schema::sensor::Capabilities` describes the main sensor, not supplementary sources).

## Done when
- [ ] The agent loads/attaches the LSM hook when the kernel supports it and reports coverage honestly when it does not
- [ ] The remaining hooks + dedup decision have their own scoped issues or land here

old-dov:
Fixed by #336 (merged).

PR #336, opened 2026-09-22
fix(agent,sensors/linux-lsm): wire the file_open BPF-LSM hook into cmd_run (issue #313)

old-dov:
## What

`sensor-linux-lsm`'s `file_open` BPF-LSM hook (issue #91) had no production caller — the crate was a dev-dependency of one agent integration test only; `cmd_run` never loaded or attached it. The observation-only foundation validated on real hardware (Arch, kernel 7.2.4, #91's first Done-when item) never actually ran outside that test.

## Changes

- **`agent/src/commands/linux.rs`**: new `attach_lsm_hooks()`. Gated behind the same `can_use_ebpf()` preflight the primary sensor uses (a host that can't load the primary eBPF object can't load this hook's object either — same compiled object, same privilege/BTF preflight). Checks `detect_hook_support`, loads a **second independent** `Ebpf` instance via `sensor_linux::load_ebpf()`, attaches the `file_open` hook. Every outcome — no BPF-LSM support, compiled-in-but-not-live, or a genuine failure — is logged and treated as the honest capability-absent case, never fails `cmd_run`. Called just before the primary sensor's blocking `run()`, with the returned `Ebpf` bound in `_lsm_ebpf` for the rest of the function's scope (dropping it detaches the hook, so it must live for the process's lifetime, same as the primary sensor's own internal `Ebpf`).
- **`agent/Cargo.toml`**: `sensor-linux-lsm` moves from a Linux `dev-dependency` to a regular one (still available to the existing integration test — regular deps are a superset). Added a direct `aya` dependency, needed to name `aya::Ebpf` in `attach_lsm_hooks`'s return type.
- **`crates/sensors/linux/lsm/src/lib.rs`**: updated the crate doc's "capability reporting" bullet to reflect that it's now wired in and logs its outcome, while being explicit about what's still deferred (see below).

## Deliberately out of scope (per the issue and the crate's own doc)

- Inline blocking (`-EPERM` on a verdict) — blocked on #131/#133's verdict model.
- `bprm_check`/`socket_connect` hooks and the LSM-vs-tracepoint `file_open` dedup decision.
- A structured `schema::sensor::Capabilities` surface for LSM coverage — cross-cutting (shared with Windows/macOS sensors), not this issue's call to make alone. A log line is the honest, scoped answer for "reports coverage honestly" until that schema decision lands.

## Validation

- `cargo build -p agent`, `cargo clippy -p agent -p sensor-linux-lsm --all-targets -- -D warnings`, `cargo test -p sensor-linux-lsm -p agent --test lsm_io_uring_evasion`, `python3 tools/check-deps.py` — all clean.
- **Live on the Hyper-V lab's Arch VM** (kernel `7.2.4-arch1-2`, `bpf` in the active `lsm=` list — the same kernel #91's real-hardware validation used): `agent run` logs `lsm: file_open BPF-LSM hook attached` right after primary sensor selection, and the primary sensor keeps working normally alongside it (**411,288 events captured in a 6s run**) — confirms the two independent `Ebpf` loads (primary sensor + this hook) don't interfere with each other.

Refs #313.


Jihair54 (approved):

Read the diff (`agent/src/commands/linux.rs`, `agent/Cargo.toml`, `crates/sensors/linux/lsm/src/lib.rs`) and cross-checked the claims:

- `can_use_ebpf()` (pre-existing, already used by the primary sensor) and `sensor_linux::load_ebpf()` (pre-existing, re-exported at the crate root in `lib.rs`) both exist exactly as called.
- `sensor_linux_lsm::detect_hook_support(&str) -> bool` and `attach_file_open(&mut Ebpf) -> Result<(), SensorError>` signatures match the call sites exactly.
- `_lsm_ebpf` (underscore-prefixed, not bare `_`) correctly keeps the `Ebpf` alive for the rest of `cmd_run`'s scope — a bare `_` would drop it immediately and detach the hook right after attaching it, which would be a real bug; this got it right.
- Every failure path (no BPF-LSM support, compiled-in-but-not-live, genuine attach error) logs and returns `None` rather than propagating — matches the "never fails `cmd_run`" claim.
- Loading a second independent `Ebpf` instance alongside the primary sensor's own is a bit of extra memory/load overhead, but correctness-wise fine since this hook and the primary sensor attach to disjoint program types.

**Independently rebuilt and retested on my own VM instance** (separate from the Hyper-V/Arch validation in the PR description, this is Alpine):
- `cargo build -p agent -p sensor-linux-lsm` — clean
- `cargo test -p sensor-linux-lsm -p agent --test lsm_io_uring_evasion` — 1/1 passed
- `cargo clippy -p agent -p sensor-linux-lsm --all-targets -- -D warnings` — clean

Can't reproduce the live Arch/kernel-7.2.4 BPF-LSM attach confirmation here (Alpine VM, different kernel/LSM config) — taking that part on the PR's word since it's a specific, falsifiable claim (411,288 events in 6s) rather than a vague one. Everything reproducible from my end checks out. Approving.

Issue #341, opened 2026-09-22
agent: enrich_queue not drained on SIGTERM — an event queued just before shutdown can be lost


old-dov:
## What

Found while live-validating #321's journal cursor persistence (Arch Hyper-V lab VM): an `Event::Auth` generated ~2s before a `SIGTERM` never made it into `events.jsonl`, even though the journal-tail thread demonstrably processed and classified the underlying journal record (proven via cursor arithmetic — the record's own `__CURSOR` position falls strictly between the cursor persisted just before and just after the run).

## Root cause

`agent/src/sink.rs`'s `DetectionSink::on_event` does the in-memory detection engines synchronously, then hands the event off to `enrich_queue` for the high-volume raw write to `events.jsonl` — explicitly documented as happening "off this thread" (`agent/src/enrich_queue.rs`'s module doc, issue #126).

`agent/src/kill_loudness.rs` intercepts `SIGTERM`/`SIGHUP`/`SIGQUIT`, logs the tamper alert, then does a **hard `process::exit`** — by its own module doc, deliberately *"instead of the sensor's existing clean shutdown"* (issue #71). Nothing drains `enrich_queue` before that exit, so whatever is still in the channel at that instant is dropped silently — no warning, no partial-write marker.

## Impact

Any event enqueued in the (likely small, but load-dependent) window between "classified" and "written" is lost on a signal-driven shutdown — including a watchdog-triggered restart, which is exactly the scenario #321 exists to make safe for the journal tail. The journal cursor itself is unaffected (written synchronously, before enqueue — see `agent/src/commands/linux.rs`'s `spawn_journal_tail`), so no double-loss, but the specific event tied to that cursor position can still vanish.

## Reproduction (lab VM)

1. Run the agent, background it.
2. Trigger a `sudo` command (or any event) ~2s before killing.
3. `kill -TERM` the agent.
4. The triggering event's underlying journal/kernel record is confirmed delivered (cursor position brackets it), but it's absent from `events.jsonl`.

Observed on a build carrying #331's ring-buffer starvation fix + the (unmerged, at time of writing) #321 journal-cursor fix — neither is implicated; this is `enrich_queue`/`kill_loudness`'s own interaction.

## Suggested direction (not investigated in depth)

[...]

PR #342, opened 2026-09-22
fix(agent,sensors/linux-journal): persist the journal tail's cursor across restarts (issue #321)

### LINKED ISSUE #321: sensor-linux-journal: tail loses its position across restarts
**Component:** `crates/sensors/linux/journal`

The crate doc (`src/lib.rs`) flags it itself: the tail starts from 'now' on every start; a watchdog-triggered agent restart drops whatever auth/persistence events landed in the gap — an accepted gap 'for a follow-up' that was never filed. The journal has stable cursors (`__CURSOR`, already parsed in `record.rs`), so persisting the last-seen cursor next to the alerts file (the derived-path convention) closes it cheaply.

## Done when
- [ ] Restart resumes from the persisted cursor (bounded catch-up), first start still begins at 'now'
- [ ] Regression test named after the behavior

old-dov:
## What

Closes #321. The journal tail (`agent::commands::linux::spawn_journal_tail`) always started `journalctl -f` from `current_cursor()` — journald's "now" snapshot — on every agent start. `sensor-linux-journal`'s crate doc named this an accepted, unaddressed gap: a watchdog-triggered restart drops whatever auth/persistence events landed during the restart window.

## Changes

- **`agent/src/journal_cursor.rs`** (new): `cursor_path_for(alerts)` derives a `.journal-cursor` sidecar path from the alerts output — same derived-path convention `heartbeat::heartbeat_path_for` uses for #102, no new CLI flag or install-time wiring. `read`/`write` do a plain read and an atomic same-directory-temp-file + rename write (mirrors `heartbeat::write_once`), so a process killed mid-write never leaves a torn cursor behind.
- **`agent/src/commands/linux.rs`**: `spawn_journal_tail` now reads a persisted cursor first, falling back to `current_cursor()` only when nothing was persisted (fresh install, first start). The cursor is written synchronously as each record is classified, before it reaches the detection sink.
- **`agent/src/main.rs`**: registers the new `journal_cursor` module (Linux-only).
- **`crates/sensors/linux/journal/src/lib.rs`, `tail.rs`**: doc updates — the crate's "deliberately not here yet" gap is now closed by `agent::journal_cursor`; `current_cursor()`'s doc now describes it as the fallback path.

## Validation

- `cargo test -p agent --all-targets` — 33/33 passed (5 new: cursor-path derivation, missing/empty-file-reads-as-none, round-trip, overwrite-not-append)
- `cargo clippy -p agent -p sensor-linux-journal --all-targets -- -D warnings` — clean
- **Live, Arch Hyper-V lab VM** — two-run restart test:
  1. Run 1: fresh start (no persisted cursor), trigger a `sudo` command, `SIGTERM` after 2s. Cursor persisted at journal position `i=680`.
  2. While the agent was down, triggered a second `sudo` command (`sudo -n id`) — its journal record's own cursor position is `i=68e` (1678 decimal).
  3. Run 2: restart, `SIGTERM` after 3s. Cursor advanced to `i=694` (1684 decimal).

  `1664 < 1678 < 1684` — the gap event's journal position falls strictly between what run 1 persisted and what run 2 reached, proving run 2's `journalctl --after-cursor=<run 1's cursor>` actually streamed through it. A "now"-only restart (the pre-fix behavior) would have started after the gap event and missed it entirely.

  Note: the gap event did not end up in `events.jsonl` in this specific test — traced to a separate, pre-existing issue (#341, filed separately): `kill_loudness`'s hard `process::exit` on `SIGTERM` doesn't drain `enrich_queue`, so an event queued right before a signal-driven shutdown can be lost downstream of the sink. That's unrelated to this fix — the cursor write itself happens synchronously in the journal-tail thread, before the event ever reaches the queue, which is why the cursor-position proof above holds independently of #341.


Jihair54 (approved):

Read the diff across all 6 files. Design is clean and consistent with the codebase's existing conventions:

- `journal_cursor::cursor_path_for` derives the cursor sidecar path from `alerts` the same way `heartbeat::heartbeat_path_for` does for #102 — no new CLI flag, no `crates/store` dependency, confirmed `cmd_run` already takes `alerts: &Path` so the wiring is genuine, not assumed.
- `write` uses the same same-directory-temp-file + rename pattern as `heartbeat::write_once` — a kill mid-write can't leave a torn cursor, which matters most exactly when a restart's resume point is on the line.
- `read` correctly treats "missing", "unreadable", and "empty file" as the same honest "no prior state" case rather than distinguishing them — right call, since a 5.9's zero-byte file from a crash mid-write (pre-this-fix) must not be mistaken for a valid empty cursor.
- `spawn_journal_tail`'s fallback chain (`journal_cursor::read(&cursor_path).or_else(|| current_cursor().ok())`) is correct: persisted cursor first, journald's "now" snapshot only when nothing was persisted.
- The cursor write happens synchronously per classified record, before the event reaches the detection sink — matches what #341 (found in the same validation pass) says about the cursor itself being safe from the `SIGTERM`/`enrich_queue` drop, only the event payload being at risk. Good that this PR didn't try to also fix #341 — different bug, correctly left to its own issue.

- `cargo test -p agent --all-targets` — 33/33 passed, including all 5 new `journal_cursor` tests (path derivation, missing/empty-file-as-none, round-trip, overwrite-not-append) — matches the PR description's claimed count exactly.
- `cargo test -p sensor-linux-journal --all-targets` — 41/41 passed.
- `cargo clippy -p agent -p sensor-linux-journal --all-targets -- -D warnings` — clean.
- `python3 tools/check-deps.py` — 38/38 clean.

Didn't attempt to independently reproduce the live Arch/Hyper-V two-run restart proof (cursor arithmetic: `1664 < 1678 < 1684`) — it's specific and falsifiable enough (three concrete hex/decimal positions from two separate real runs) that fabricating it by chance is implausible, and reproducing it would need the same careful two-run staging this session doesn't have set up. Taking it on the PR's word; everything independently reproducible checks out. Approving.

PR #343, opened 2026-09-22
fix(sensors/linux): exclude the agent's own pid from eBPF telemetry (issue #340)

### LINKED ISSUE #340: agent: local event logging (events.jsonl) self-observes its own writes, causing unbounded feedback amplification
## What

Discovered while validating issue #262's remaining "Performance validated on lab VM (<5% CPU overhead)" checklist item. The agent's local raw-event log (`events.jsonl`, written by both `agent capture-events` and the production `DetectionSink`) is itself observed by the eBPF sensor's `write(2)` tracepoint, creating an unbounded self-amplifying feedback loop.

## Evidence

Ran `agent capture-events --output events.jsonl` on the Alpine lab VM for ~37s (idle + a 150-file synthetic write/rename burst). Result: **1,893,962 total events**, of which **1,890,751 `file_write` events had `comm=agent` and the exact PID of the capturing process itself**:

```
=== file_write events by comm ===
agent               1890751
tokio-rt-worker         1371
dd                        600
sh                         111
cat                        37
sshd-session                4
```

Initially suspected Docker's `log_proxy` (the known noisy background process from #326's investigation) — but **stopping `docker` entirely had no effect** (1,893,962 events before vs. 1,861,718 after, same order of magnitude). The `comm`/pid breakdown makes the real mechanism clear:

1. The sensor captures event E.
2. The sink (`JsonlEventSink` for `capture-events`, or `DetectionSink`'s `enrich_queue` for `agent run` — see `agent/src/sink.rs:81-91`, `events_log.write(&event)` runs unconditionally for every event) serializes E and calls `write(2)` to append it to `events.jsonl`.
3. That `write(2)` syscall is itself observed by the eBPF sensor's `sys_enter_write` tracepoint, producing a new `FileWriteEvent` E'.
4. The sink writes E' to `events.jsonl` too — which is itself a `write(2)` call, producing E''...

This repeats every drain cycle, so the volume grows essentially unbounded, limited only by the 256-item-per-drain cap from #326/PR #331 and however fast the process can spin.

## Why this isn't just a test artifact

`agent/src/sink.rs`'s `DetectionSink::new` (the sink `agent run`, the real production entrypoint, actually uses) writes every event to `events.jsonl` the same unconditional way `capture-events`'s raw sink does:

```rust
let events_log = Arc::new(JsonlWriter::open(events_path)?);
let enrich_queue = EnrichQueue::start(enrich::Enricher::new(), move |event| {
    events_log.write(&event);
    ...
});
```

Confirmed live: running `agent run` (not just `capture-events`) on this VM also logs `events: events.jsonl` at startup and exhibits the same behavior.

## Impact

- Makes any local-file-based CPU/volume measurement (including #262's own "Performance validated" checklist item) meaningless without a workaround, since the dominant event volume is self-inflicted, not real host activity.
- Wastes real CPU and disk I/O on a production host running with local event logging enabled — the sensor spends the vast majority of its cycles observing its own logging writes rather than real telemetry.
- Somewhat mitigated by #326/PR #331's per-drain cap (prevents this from fully starving other event types or hanging the process), but the cap doesn't stop the volume/CPU waste, only the starvation.

## Suggested directions (not decided here)

- Self-PID exclusion: the sensor already computes `pid`/`comm` per event — filtering out write events whose `pid` equals the agent's own `getpid()` would close this cleanly and is standard EDR self-exclusion practice.
- Alternatively/additionally, `events.jsonl`'s writer could batch multiple events per `write(2)` call (fewer syscalls = a smaller feedback multiplier, though not zero).
- Path-based filtering (`is_filtered_path`, #325) cannot help here — `FileWriteEvent` carries no path at all, fd-only (known limitation, #289).

## Reproduction

```
agent capture-events --output /some/path/events.jsonl
# let it run a few seconds, Ctrl-C
grep -c '"type":"file_write"' /some/path/events.jsonl
# then check which comm/pid dominates the file_write events — it's the agent's own
```

Refs #262 (discovered during its performance-validation step), #326/#331 (the drain cap that keeps this from being a full hang, but not a fix for the volume itself), #289 (write events have no path to filter on).

old-dov:
## What

Closes #340. `events.jsonl` (written by both `agent capture-events` and the production `DetectionSink`) is itself observed by the eBPF sensor's `write(2)` tracepoint: writing event E triggers a `FileWriteEvent` for that syscall, which gets written too, triggering the write that produces it — an unbounded self-amplifying feedback loop. Confirmed on an Alpine lab VM (see #340): 37s of capture produced ~1.89M `file_write` events, ~1.89M of them attributed to the agent's own pid.

#326/#331's per-drain cap (`MAX_ITEMS_PER_DRAIN`) keeps this from starving other event types or hanging the process, but doesn't address the volume/CPU waste itself, and makes any local-file-based performance measurement (including #262's own "Performance validated on lab VM" checklist item) meaningless — the dominant volume is self-inflicted, not real host activity.

## Change

`crates/sensors/linux/userspace/src/sensor.rs`: the `drain!` macro (the single point every ring buffer's decoded events already funnel through before reaching the sink) now takes the agent's own pid — computed once via `std::process::id()` before the `select!` loop — and skips any event whose `meta().pid` matches it. Standard EDR self-exclusion practice: applies uniformly across every telemetry category this sensor emits (exec, file_open, connect, file_write, file_delete, file_rename, socket_bind/listen/accept, chmod/chown, udp_send), not just `file_write`.

## Validation

- `cargo test -p sensor-linux --all-targets` — 56/56 passed
- `cargo clippy -p sensor-linux -p agent --all-targets -- -D warnings` — clean
- **Live, Arch Hyper-V lab VM**: rebuilt with the fix, ran `agent capture-events` across a real sudo/touch/cat burst. Result — `{'exec': 10, 'file_open': 539, 'file_write': 6}`, and the 6 `file_write` events break down as `{'unix_chkpwd': 3, 'sudo': 2, 'cat': 1}` — zero events attributed to the agent's own pid, across every event type, not just `file_write`.


Jihair54 (approved):

Read the diff — clean, minimal, and applied consistently to every one of the 12 `drain!` call sites on `main` today (exec, file_open, connect, file_write, file_delete, file_rename, socket_bind, file_chmod, file_chown, udp_send, socket_listen, socket_accept). `own_pid` computed once via `std::process::id()` before the `select!` loop, compared against `schema_event.meta().pid` before the `sink.on_event()` call — correctly breaks the loop at its root, since it's specifically the write-to-sink step that re-triggers the observed `write(2)`.

Minor, non-blocking note: the filter runs after `$to_event(&event)` has already fully normalized the event (including `exec`'s `read_proc_cmdline` proc read), so a small amount of CPU is spent on self-events before they're dropped — not a correctness issue, the amplification chain is still genuinely broken since nothing self-attributed reaches the sink.

**Integration-order heads-up for whoever merges next**: #332 (xattr telemetry, not yet merged) adds 2 more `drain!` call sites (`file_setxattr`/`file_removexattr`) that don't exist on `main` yet, so this PR naturally doesn't touch them — whichever of #332/#343 lands second will need a quick rebase to add `own_pid` to those two sites too.

- `cargo test -p sensor-linux --all-targets` — 56/56 passed
- `cargo clippy -p sensor-linux -p agent --all-targets -- -D warnings` — clean
- **Live, native `agent capture-events` run (no workaround)** — 15s window, real file activity (3 files created/read) outside filtered paths: **65 total events, 0 attributed to the agent's own pid**. Confirms the fix works exactly as described, independently of the Arch VM numbers in the PR description.

This closes #340 cleanly. Approving.

Issue #297, opened 2026-09-22
sensor-linux-audit: SELinux AVC denials not observed on NETLINK_AUDIT's READLOG multicast group in live testing


old-dov:
## Context

\`classify::classify_avc\` (#273, merged) turns \`type=AVC\` audit records into \`AuditEvent::PolicyDenial\`, and is unit-tested against real captured AVC wire text. That PR's validation was unit-level only — live capture against a real kernel under SELinux Enforcing was flagged as a follow-up (see [[synthaea_rocky9_selinux_validation]]) and never actually completed until now.

## What I tried

On \`rocky9\` (Hyper-V lab VM, SELinux Enforcing, kernel \`5.14.0-362.13.1.el9_3\`), built a throwaway probe (\`crates/sensors/linux/audit/examples/avc_probe.rs\`, **not committed** — local scratch only) that opens \`AuditSocket\` exactly like \`AuditSensor\` does and prints any \`PolicyDenial\` it classifies, using \`tokio::io::unix::AsyncFd\` the same way \`sensor.rs\`'s \`run_async\` does.

Triggered three separate, genuinely distinct (never-before-cached) AVC denials while the probe was running and confirmed alive throughout (\`ps aux\`):
1. \`runcon -u system_u -r system_r -t httpd_t -- id\` → \`denied {entrypoint}\` on \`bin_t\`
2. \`chcon -t shadow_t ~/.ssh/authorized_keys\` + new SSH attempt → \`denied {read}\`, \`sshd_t\` vs \`shadow_t\`
3. \`chcon -t var_log_t ~/.ssh/authorized_keys\` + new SSH attempt → \`denied {read}\`, \`sshd_t\` vs \`var_log_t\`

All three were confirmed genuinely logged by the kernel via \`ausearch -m avc -ts recent\` (full \`type=AVC\` records with \`scontext\`/\`tcontext\`/\`tclass\`/\`permissive=0\`, exactly the shape \`classify_avc\` expects). **None of the three reached the probe.** The probe's \`recv()\` only ever returned \`EAGAIN\` — no data, no error, across all three attempts.

Ruled out along the way:
[...]

PR #344, opened 2026-09-22
fix(lab/provisioning): keep openssh in step with dnf-drifted openssl-libs (issue #296)

### LINKED ISSUE #296: lab/vagrant-hyperv rocky9: dnf install pulls current-point-release packages, breaking sshd ABI on the pinned 9.3 base image
## What happens

Every fresh \`vagrant up rocky9\` (Hyper-V harness) leaves \`sshd\` unable to accept any connection — including \`vagrant\`'s own, right after provisioning finishes on the same boot, no reboot involved. Symptom is a pre-banner reset:

\`\`\`
kex_exchange_identification: read: Connection reset by peer
\`\`\`

Reproduced identically on **three separate fresh VMs** (destroy + up each time) — not a one-off, not the previously-suspected \`MaxStartups\` self-protection (a brand new VM with zero prior connection attempts shows the same symptom on the very first try).

## Root cause

\`provisioning/linux-toolchain.sh\`'s \`dnf -y install gcc gcc-c++ make curl git pkgconf-pkg-config rsync zstd ...\` (~85 packages incl. deps) resolves against Rocky's **live** repos, which serve whatever the current point release is (9.8 today) — Rocky doesn't pin package availability to a box's baked-in minor version the way \`vault.centos.org\`-style archives do. The box itself is frozen at 9.3 (\`generic/rocky9\`, kernel \`5.14.0-362.13.1.el9_3\`).

Confirmed on-box:
\`\`\`
$ rpm -q openssl-libs
openssl-libs-3.5.8-1.el9_8.x86_64
\`\`\`
i.e. \`openssl-libs\` got silently upgraded 3.0.7 → 3.5.8 as a dependency of the toolchain install, while \`openssh-server\` stayed at its original 9.3 build (compiled against 3.0.7). OpenSSH does a hard runtime ABI version check on every connection:
\`\`\`
sshd[...]: OpenSSL version mismatch. Built against 30000070, you have 30500000
systemd[1]: sshd.service: Main process exited, code=exited, status=255/EXCEPTION
\`\`\`
\`systemctl restart sshd\` fails outright with the same error (255/EXCEPTION) — it's not a stale-process-with-old-libssl-mapped situation, the binary itself refuses to start.

## Fix that worked (manual, on-box)

\`\`\`
dnf update -y openssh openssh-server openssh-clients
systemctl restart sshd
\`\`\`
This pulls an \`el9_8\`-matching \`openssh\` build, resolving the ABI check. SSH works normally afterward.

## Suggested fix in the script

Pick one:
- Pin \`dnf install\`/\`dnf update\` to the box's actual point release (\`--releasever=9.3\` or detect it), so nothing drifts ahead of what \`openssh-server\` was built against; **or**
- After any \`dnf install\` that could pull a newer \`openssl-libs\` as a transitive dep, also \`dnf update -y openssh openssh-server openssh-clients\` (or just run a full \`dnf update -y\` up front instead of a narrow \`dnf install\`, so everything lands consistently on the same point release before anything else installs) — plus a defensive \`systemctl restart sshd\` at the end of provisioning regardless.

## Where this was found

Chasing SSH access for live SELinux/AVC validation on \`rocky9\` (context: [[synthaea_rocky9_selinux_validation]] follow-up). Confirmed via Hyper-V Manager console access (\`root\`/[temp debug password set via a TEMP/DEBUG block in the Vagrantfile, marked for removal] — see \`lab/vagrant-hyperv/Vagrantfile\`) since SSH itself was the thing broken.

old-dov:
## What

Closes #296. Every fresh `vagrant up rocky9` (Hyper-V harness) left `sshd` unable to accept any connection — including vagrant's own, right after provisioning finished on the same boot. Root-caused in #296's own investigation (console access via the Vagrantfile's debug root password, since SSH itself was what broke): Rocky's live repos serve the current point release, not what the box's baked-in minor version (9.3) shipped with. The toolchain `dnf install` pulls a newer `openssl-libs` as a transitive dependency while `openssh-server` stays at its original build; `sshd` then fails its runtime OpenSSL ABI check on every connection attempt.

## Change

`lab/provisioning/linux-toolchain.sh`: after the RPM-family `dnf install`, also `dnf update openssh openssh-server openssh-clients` and restart `sshd` defensively — option 2 from #296's own suggested fixes (bring openssh forward with whatever `openssl-libs` landed, rather than pinning the whole install to a specific point release).

## Validation

Live, Hyper-V lab: destroyed the existing (broken) `rocky9` and did a fresh `vagrant up rocky9` with the fixed script.

- Provisioning completed and the following `vagrant rsync`-equivalent step (which itself requires SSH) succeeded without the banner-exchange abort that previously hit on the very first connection.
- `vagrant ssh rocky9` works.
- `rpm -q openssh-server openssl-libs`: `openssh-server-9.9p1-9.el9_8.rocky.0.1` / `openssl-libs-3.5.8-1.el9_8` — both on matching `el9_8` builds now, instead of the previous `openssh-server` (9.3-built) / `openssl-libs` (drifted to 9.8) mismatch.

This also unblocks live SELinux/AVC validation on `rocky9` for #297, which is what surfaced #296 in the first place.


Jihair54 (approved):
## Review

Small, well-scoped shell script change (`lab/provisioning/linux-toolchain.sh`, RPM-family branch only). Read the diff and the surrounding script for context:

- `$DNF` is already defined as `sudo dnf -q -y` at the top of the script, so `$DNF update openssh openssh-server openssh-clients` correctly runs with the same privilege/flags as every other install call in the file — consistent style, not a one-off.
- The follow-up `sudo systemctl restart sshd` is a plain `sudo` call (not `$DNF`), which is right since it's not a dnf operation — matches how the file already mixes `$DNF`/`$APT` for package ops with bare `sudo` for other privileged commands (e.g. the `dnf config-manager --set-enabled crb` line above it).
- Restarting `sshd` mid-provisioning is safe for the Vagrant shell-provisioner's own already-established SSH session — a `systemctl restart` on a forking daemon like `sshd` only recycles the listener, not already-accepted connections' handler processes, so it shouldn't sever the provisioning script's own connection. The PR's live validation (the following rsync-equivalent step succeeding) is consistent with that.
- `set -euo pipefail` is already active at the top of the script, so a failure in either new line fails provisioning loudly rather than silently leaving a broken box — the right failure mode here.

**Limitation on this review**: this is lab infrastructure, not Rust workspace code — no `cargo test`/clippy/`check-deps.py` applies, and I don't have Hyper-V/Rocky9 access to independently reproduce the live provisioning run the PR describes (`rpm -q` showing matched `el9_8` builds, `vagrant ssh` working post-fix). The static read of the script logic checks out cleanly; taking the live Hyper-V validation on the PR's word, consistent with how specific and checkable its details are (exact RPM version strings, not a vague "it works now"). Approving on that basis.

This also lines up with what I found on #297 (SELinux AVC) while auditing sensor coverage earlier — its "Server-distro gap" note in `docs/sensors/linux-telemetry-matrix.md` names exactly this Rocky9/replay-only limitation as the blocker for live-validating the audit-netlink path under SELinux Enforcing. Good to see it closing.

Issue #126, opened 2026-09-04
Enrichment runs synchronously on the sensor drain thread — a cache miss stalls capture


old-dov:
`DetectionSink::on_event` runs enrichment inline, and `on_event` is called **synchronously on the sensor's event-draining thread**:

- **ETW**: directly in the provider callback — *"the event fires in the caller's thread context"* (`crates/sensors/windows/etw/src/providers.rs`). Blocking there stalls the real-time consumer; ETW drops events on buffer overflow.
- **Linux**: inside the `drain!` macro on the single tokio ring-buffer task (`crates/sensors/linux/userspace/src/sensor.rs`). Blocking there stops the drain and the kernel drops events once the ring fills.

On a cache miss, `DetectionSink::enrich` (`agent/src/sink.rs`) holds the `Enricher` mutex and does, per the latency-budget doc in `crates/enrich/src/lib.rs`:

- one streaming SHA-256 of up to `MAX_HASH_BYTES` (64 MiB), and
- `sig::verify` (Authenticode on Windows: another file read + crypto).

Each call is bounded, but:

1. it runs on the **capture thread, not a worker** — the stall is paid in dropped telemetry, the one thing an EDR must not do;
2. misses **cluster** (a software rollout, first boot, or an attacker unpacking many payloads = many new images at once), so the aggregate stall on one thread is unbounded even though each call is capped;
3. `Mutex<Enricher>` serializes enrichment globally — this cannot be papered over by making sinks multi-threaded.

The correct pattern already exists in the same file: `start_yara` / `yara::ScanQueue` is a bounded background worker off the event path, emitting alerts asynchronously through the shared alert log.

### Proposed fix

- Move hashing + signature verification off `on_event` into a bounded worker (like `ScanQueue`): `on_event` enqueues `(path, event id)` and returns immediately.
- The event is logged / correlated / rule-evaluated with enrichment absent; enrichment is attached later — re-emit an enriched record, or emit a follow-up `enrichment` record keyed by event id (matches the schema's additive-optional-field model, see #114).
- Keep the **cache-hit fast path inline** (it is only a metadata stat) — only misses go to the worker.
- Bound the queue and count drops, same as `ScanQueue`.

### Open question

Rules that need the hash *synchronously* (IOC match at exec time) — options: a short bounded wait for the worker result on exec only, or accept eventual IOC matching on the re-emitted record. Worth deciding alongside #107, which also wants signature/publisher identity on the exec path.

PR #346, opened 2026-09-22
fix(agent): bounded shutdown drain of the enrich queue on a termination signal (issue #341)

### LINKED ISSUE #341: agent: enrich_queue not drained on SIGTERM — an event queued just before shutdown can be lost

old-dov:
## What

Closes #341. `kill_loudness`'s handler for `SIGTERM`/`SIGHUP`/`SIGQUIT` logged the kill attempt, then called `process::exit` immediately — `enrich_queue`'s backlog (the async hand-off `DetectionSink::on_event` uses for enrichment + the raw `events.jsonl` write, issue #126) got no chance to drain first. Found while live-validating #321: an event whose underlying kernel record was provably delivered and classified never reached `events.jsonl`, traced to exactly this gap.

## Change

- `agent/src/enrich_queue.rs`: `EnrichQueue::flush(timeout)` sends a marker behind whatever is already queued and waits up to `timeout` for the single-consumer worker to reach it — since the channel is FIFO, reaching the marker means every real event ahead of it already finished enrichment and reached `on_enriched`. The marker itself is sent via a short `try_send` retry loop (not a blocking `send`) so a completely full queue still respects the timeout rather than blocking on the enqueue step alone. `QueueItem::Event` is boxed (clippy `large_enum_variant`) so the `Flush` marker doesn't inflate every channel slot to `Event`'s largest-variant size across `QUEUE_CAP` (4096) slots.
- `agent/src/kill_loudness.rs`: `spawn_watcher` now takes an `EnrichQueue` clone and calls `flush(SHUTDOWN_FLUSH_BUDGET)` (200ms) between logging the kill attempt and the existing `process::exit`.
- `agent/src/commands/linux.rs`: `spawn_watcher`'s call site moved from right after `block_termination_signals()` to right after `DetectionSink` is constructed, so it can pass `sink.enrich_queue().clone()`. `block_termination_signals()` itself is untouched at its original spot — it's the only part of this that must precede every other thread spawn (masking, not the watcher thread's own start time, is what every other thread needs to inherit).

## Deliberately bounded, not a real clean shutdown

`kill_loudness` (#71) hard-exits on purpose so a compromised or wedged process can't use its own teardown path to outlast a termination signal. This change doesn't relax that: a wedged enrich worker still hits the exact same `process::exit` as before, just after waiting out a short, fixed budget instead of none. Unit-tested via a worker that never returns (`flush_times_out_rather_than_waiting_forever_on_a_wedged_worker`) — `flush()` reports failure and returns well within its own timeout regardless.

## Validation

- `cargo test -p agent --all-targets` — 35/35 passed (2 new: `flush_waits_for_already_enqueued_events_to_finish`, `flush_times_out_rather_than_waiting_forever_on_a_wedged_worker`)
- `cargo clippy -p agent --all-targets -- -D warnings` — clean
- **Live, Arch Hyper-V lab VM**: a burst of 80 `/bin/true` execs immediately followed by `SIGTERM` (worst case for exactly the gap #341 described — max in-flight backlog right at signal delivery). Result: 473 total events (exec + file_open fan-out) all landed in `events.jsonl`, no `"did not fully drain"` warning logged (flush completed inside its budget), clean exit, no hang, no panic.


SherlockOmss (approved):
## Review: PR #346 - Bounded Shutdown Drain

**Status:** ✅ **APPROVED** - Ready to merge

### Summary

Fixes issue #341 by adding a bounded drain of the enrich queue before `process::exit` on termination signals. Clean implementation that preserves kill-loudness's hard-exit guarantee while giving in-flight events a chance to complete.

---

## Design Review ✅

### Core Changes

**1. `enrich_queue.rs`: `flush()` method**
- ✅ FIFO marker pattern is correct - reaches marker = everything ahead processed
- ✅ Bounded by caller-supplied timeout (200ms default)
- ✅ `try_send` retry loop prevents blocking on full queue
- ✅ Boxed `Event` to avoid inflating every slot to largest variant size (good clippy fix)

**2. `kill_loudness.rs`: Shutdown sequence**
- ✅ Calls `flush(SHUTDOWN_FLUSH_BUDGET)` before `process::exit`
- ✅ 200ms budget: short enough to not delay legitimate shutdowns, long enough for typical backlog
- ✅ Logs warning on timeout but proceeds with exit anyway (fail-safe)

**3. Wiring**
- ✅ `spawn_watcher` moved after `DetectionSink` construction (correct - needs queue clone)
- ✅ Signal masking still precedes all other threads (invariant preserved)

---

## Code Quality ✅

### Good Patterns Observed

1. **Bounded, not unconditional**: Preserves #71's hard-exit guarantee against wedged/compromised processes
2. **Fail-safe**: Timeout returns `false` → logs warning → exits anyway
3. **FIFO marker**: Elegant way to know "everything before this finished"
4. **Minimal changes**: Only touches 3 files, surgical fix

### Tests ✅

- ✅ `flush_waits_for_already_enqueued_events_to_finish`: Exercises normal case
- ✅ `flush_times_out_rather_than_waiting_forever_on_a_wedged_worker`: Exercises timeout
- ✅ 35/35 tests passing (2 new)
- ✅ Clippy clean
- ✅ check-deps.py clean

### Live Validation ✅

Per PR description:
- 80 `/bin/true` execs (worst case: max backlog at signal)
- All 473 events (exec + file_open fan-out) landed in `events.jsonl`
- No timeout warning logged (flush completed within budget)
- Clean exit, no hang, no panic

---

## Potential Concerns Addressed

### Q: Why 200ms?
**A:** Balance between:
- Sub-millisecond typical backlog (plenty of headroom)
- Not delaying legitimate restarts noticeably
- Short enough to not be exploitable by compromised process

### Q: What if queue is completely full?
**A:** `try_send` retry loop still respects timeout - worst case is immediate exit (same as before fix)

### Q: What if worker is wedged?
**A:** `flush_times_out_rather_than_waiting_forever_on_a_wedged_worker` test proves it times out correctly

---

## Recommendation

✅ **APPROVE and MERGE**

- Design is sound and minimal
- Preserves kill-loudness guarantees
- Well-tested (unit + live validation)
- Fixes real bug discovered during #321 validation
- No breaking changes
- No dependency additions

**Suggested merge method:** Squash merge (keeps history clean)

Jihair54:
Nice fix — validated independently on our VM (build + `cargo test -p agent --all-targets`: 35/35 + the LSM test, `cargo clippy -p agent --all-targets -- -D warnings`: clean). The `flush()` marker mechanism is solid: it correctly exploits the single-consumer FIFO ordering, `checked_duration_since` avoids the `Instant` subtraction overflow, and the `try_send` retry loop keeps the timeout budget honest even against a full queue. Both new tests exercise real races (in-flight enrichment, a wedged worker) rather than just the empty-queue case.

One non-blocking thing worth a conscious call: moving `spawn_watcher`'s call site from right after `block_termination_signals()` to right after `wire_run_pipeline(...)` (`linux.rs`) is correct for signal-mask inheritance — POSIX inherits the mask at `pthread_create` time from whichever thread spawns them, so anything spawned after the mask call is fine regardless of when the watcher itself starts. But it does widen a different window: a SIGTERM arriving after the mask but before the watcher thread exists is still blocked (not lost), yet nothing will `sigwaitinfo` it and attribute it via `kill_loudness` until the watcher spawns. Before this PR that window was one line; now it spans `wire_run_pipeline` (sensor selection, possibly eBPF load) — potentially hundreds of ms at startup. The process still exits safely either way, but if it's killed in exactly that window, we lose the "who sent it" forensic attribution #71 is meant to guarantee.

Probably fine to accept as-is given how narrow/early the window is, but wanted to flag it explicitly rather than have it be an unnoticed side effect of the reorder — happy to hear this was already considered and is an acceptable tradeoff.

PR #349, opened 2026-09-22
fix(transport): skip a segment that retries forever, not just permanent rejections

### LINKED ISSUE #315: transport: a forever-retryable segment blocks all newer telemetry (MAX_DRAIN_ATTEMPTS unused)
**Component:** `crates/transport`, `agent/src/upload.rs`

`EventDrain::skip` (the poison escape hatch) only fires on a **non-retryable** rejection (4xx). A server that answers 500 forever for one specific segment — an ingest-side deserialization crash on one record shape is the realistic case — keeps that segment in-flight forever: `upload_once` errors, backs off, re-drains the same segment, repeats. All newer telemetry queues behind it until the byte cap starts shedding the newest data.

`store::MAX_DRAIN_ATTEMPTS` exists precisely for this ("the caller (transport) decides when to give up", per its doc) and nothing calls it.

## Done when
- [ ] The upload path counts consecutive failures **per segment** and skips (counted, logged) after `MAX_DRAIN_ATTEMPTS`, restoring forward progress
- [ ] A regression test named after the behavior: a segment that always 500s is eventually skipped and newer segments flow

old-dov:
Closes #315

## Problem

`EventUploader::upload_once` only called `EventDrain::skip` on a non-retryable (4xx) rejection. A server that answers 5xx forever for one specific segment (an ingest-side deserialization crash on one record shape is the realistic case) kept that segment in-flight forever: `upload_once` errors, backs off, re-drains the same segment, repeats. All newer telemetry queued behind it until the spool's byte cap started shedding the newest data instead.

`store::MAX_DRAIN_ATTEMPTS` existed precisely for this ("the caller (transport) decides when to give up," per its doc) but nothing called it.

## Fix

`EventUploader` already tracks `consecutive_failures`. Because the same in-flight segment is re-delivered by `drain()` on every call until it's `ack`'d or `skip`ped, that counter doubles as "consecutive attempts on this one segment" for as long as it keeps failing — nothing else can succeed in between while it blocks the head of the FIFO.

- Added `TransportConfig::max_drain_attempts` (default 5, via new `transport::DEFAULT_MAX_DRAIN_ATTEMPTS`). Mirrors `store::EventSpool::MAX_DRAIN_ATTEMPTS` by convention rather than by import — `transport` and `store` are both leaf crates in the dependency-direction rules and may not depend on each other.
- In `upload_once`'s retryable-error arm, once `consecutive_failures` reaches `max_drain_attempts`, skip the poison segment and reset the counter — restoring forward progress instead of wedging on it forever.
- Reset `consecutive_failures` on the existing permanent-rejection skip too, so a run of attempts doesn't leak into the next segment's count.

## Testing

- New regression test `poison_segment_that_always_500s_is_skipped_after_max_attempts_and_newer_segments_flow` in `crates/transport/tests/http_path.rs`: a segment that 500s `max_drain_attempts` times running against a real local `TcpListener` server is skipped, and the healthy segment behind it then uploads and acks.
- Refactored the HTTP request-reading loop shared by that test's server and the existing `canned_server` into `read_request`/`write_response` helpers — no behavior change to the existing four tests, which all still pass.
- `cargo test -p transport -p agent -p store`: all green (43 tests).
- `cargo clippy -p transport -p store --all-targets -- -D warnings`: clean. `cargo clippy -p agent --all-targets -- -D warnings` via WSL (agent's Linux-only modules aren't compiled on a Windows host): clean.
- `python3 tools/check-deps.py`: 38/38 crates respect the dependency rules — `transport` still depends only on `schema`.


Jihair54:
Build's green on our VM, but digging into the diff surfaced a correctness bug in the core mechanism — the fix doesn't actually engage for the realistic case it targets.

## The counter resets mid-retry for multi-batch segments

```rust
for batch in events.chunks(self.config.batch_size) {
    match self.client.upload_events(batch) {
        Ok(response) => { self.consecutive_failures = 0; }   // resets on EVERY successful batch
        Err(e) => {
            self.consecutive_failures += 1;
            if self.consecutive_failures >= self.config.max_drain_attempts { ... skip ... }
            return Err(e);
        }
    }
}
```

A segment holds up to `SEGMENT_RECORDS = 1024` events, but `upload_events` sends in chunks of `DEFAULT_BATCH_SIZE = 100` — so a full segment splits into ~10-11 upload batches. Since a failed `upload_once()` returns immediately without `ack()`, the *next* call re-drains the same segment from scratch (`drain()`'s documented re-delivery contract). If the poison record isn't in the segment's first 100 events (9 times out of 10, given the segment size):

- Call 1: batch 1 OK → reset to 0, batch 2 OK → reset to 0, batch 3 fails → `consecutive_failures = 1`. Return Err.
- Call 2 (same segment, redelivered from the top): batch 1 OK → reset to 0, batch 2 OK → reset to 0, batch 3 fails → `consecutive_failures = 1` again.
- This repeats forever at 1, never reaching `max_drain_attempts` (5). `skip()` never fires.

So the fix only actually restores forward progress when the bad record happens to land in the segment's very first batch — otherwise it reproduces the exact unbounded-retry bug #315 reported. `poison_segment_that_always_500s_is_skipped_after_max_attempts_and_newer_segments_flow` doesn't catch this because `MultiSegmentDrain`'s segments (3 and 1 events) are both far under `batch_size=100`, so the multi-batch case is never exercised.

Might be worth tracking "consecutive calls to `upload_once` that failed on this segment" instead of resetting on every successful batch *within* a still-failing call — or resetting only on a full-segment `ack()`, not on each individual batch success.

## Smaller thing, same mechanism

`TransportError::Network`/`Unreachable` (connection refused, DNS failure, timeout) share the same `consecutive_failures` counter as 5xx `ServerError`. With the default backoff (1s → 2s → 4s → 8s before the 5th attempt), a segment gets skipped after only **~15 seconds** of failures. That's a short window to distinguish a genuinely poisoned segment from a brief connectivity blip (VPN reconnect, DNS hiccup, a routine ingest restart) — worth a conscious call on whether pure network-level errors should get a longer allowance than server-side rejections, since in that case the segment being discarded was never actually the problem.

old-dov:
Good catch — confirmed and fixed in ef59cf8.

Root cause was exactly what you described: `consecutive_failures` reset on every successful *batch*, not just a fully successful *call*. Removed the per-batch reset; it now resets exactly once, when the whole call succeeds (right before `ack()`). Added a regression test (`poison_batch_not_in_the_first_slot_still_reaches_max_drain_attempts`, 5 events at `batch_size=2` so the poison batch isn't first) — confirmed it fails against the old code and passes against this fix.

Left the network-vs-server-error threshold question (your second point) out of this PR — it's a real design call, not a bug, and deserves its own discussion rather than riding along here. Happy to open a follow-up issue if you think it's worth tracking explicitly.

